PR #3, opened 2020-04-22
Scanner

jared-nishikawa:
Some base code hacked up that we can play around with.

The scanner tokenizes input into SGF-specific tokens.
For example: `(;GM[1]FF[4]CA[UTF-8]AP[CGoban:3]ST[2]` becomes
`["(", ";", "GM", "[", "1", "]", "FF", "[", "4", "]", "CA", "[", "UTF-8", "]", "AP", "[", "CGoban:3", "]", "ST", "[", "2", "]"]`

The parser takes these tokens and parses them into nodes and branches. Nodes start with a semicolon ";" and branches start with a left paren "(".  Currently the parser just stores all these nodes in a map, indexed by an int.  Eventually, we will want to update this data structure to retain the tree structure of the game variations.


sgf/parser.go
```diff
@@ -0,0 +1,174 @@
+package sgf
+
+import (
+	"errors"
+	"io"
+)
+
+// A node is comprised of key-value pairs (might hold other info someday)
+type Node struct {
+	Fields map[string]string
```
artasparks: Typically, I've called these Properties.

Should be a list of string since SGF properties can be repeated.

Also, I would expect a Node to be a recursive type:

Children []*Node

sgf/parser.go
```diff
@@ -0,0 +1,174 @@
+package sgf
+
+import (
+	"errors"
+	"io"
+)
+
+// A node is comprised of key-value pairs (might hold other info someday)
+type Node struct {
```
artasparks: Let's move to game/

sgf/parser.go
```diff
@@ -0,0 +1,174 @@
...
+
+func NewNode() *Node {
+	n := &Node{}
+	n.Fields = make(map[string]string)
+	return n
+}
+
+// A game record (sgf) struct for now is just a collection of nodes
+// TODO: this struct should retain the branch structure of game variations
+type Sgf struct {
```
artasparks: Let's move this type to game/game.go (and call it Game)

sgf/parser.go
```diff
@@ -0,0 +1,174 @@
...
+	Index int
+	Nodes map[int]*Node
```
artasparks: Typically, I would expect a Game to have a single Root node.

sgf/parser.go
```diff
@@ -0,0 +1,174 @@
+package sgf
```
artasparks: Probably package for parser should be parse or parser.

artasparks (commented):
Couple of initial comments

sgf/scanner.go
```diff
@@ -0,0 +1,160 @@
...
+	Raw  string
+}
+
+// Scanner wraps bufio.Reader but has the public Scan() function to scan tokens
+type Scanner struct {
+	*bufio.Reader
+}
+
+// Creates a new scanner object out of an io.Reader interface
+func NewScanner(r io.Reader) *Scanner {
```
artasparks: It's probably helpful to have a method 

FromString(s string) {

}

Or honestly, just a Parse(s string) method is probably more helpful.

sgf/scanner.go
```diff
@@ -0,0 +1,160 @@
...
+		return &Token{
+			Type: RightBracket,
+			Raw:  "]",
+		}
+	case ';':
+		return &Token{
+			Type: Semicolon,
+			Raw:  ";",
+		}
+	default:
```
artasparks: Actually, wouldn't this be an error/unexpected token?

artasparks (commented):
Might be helpful: https://github.com/Kashomon/bodoko/blob/master/core/sgf_parser.go. A previous parser I wrote in go.

sgf/parser.go
jared-nishikawa: Sounds good, I've redone Node to have both children and parent (to traverse either direction, if needed)

sgf/parser.go
jared-nishikawa: Done

sgf/parser.go
jared-nishikawa: Also moved everything to

internal/
-  game/
-  parser/
-  scanner/

sgf/parser.go
jared-nishikawa: Agreed, fixed

sgf/parser.go
jared-nishikawa: I made an attempt at cleaner packaging by making scanner, parser, and game directories inside of "internal"

sgf/scanner.go
jared-nishikawa: Hmmmm... I put in FromString and FromReader for flexibility.  Are you saying we should scrap FromReader?

sgf/scanner.go
jared-nishikawa: Agreed, fixed

sgf/parser.go
jared-nishikawa: I also didn't realize the properties can be repeated.  Changed to `map[string][]string`

internal/game/game.go
```diff
@@ -0,0 +1,27 @@
+package game
```
artasparks: I don't think Internal is right for game. I think the goal should be to provide generally useful go libraries -- and certainly Game would be one of those. So, I'd make it a publicly available library at the top level.

internal/parser/parser.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return g, nil
+}
+
+// parse, for example "PW[Player White]" into
+// field="PW"
+// value="Player White"
+func (p *Parser) parseFieldValue() (string, string, error) {
```
artasparks: How does this handle something like AW[pa][qa][nb][ob]?

internal/parser/parser.go
```diff
@@ -0,0 +1,164 @@
...
+// scanSkipWhitespace is a wrapper around scan() that skips whitespace
+func (p *Parser) scanSkipWhitespace() *scanner.Token {
+	tok := p.scan()
+	if tok.Type == scanner.Whitespace {
+		tok = p.scan()
+	}
+	return tok
+}
+
+// Parses the root branch
```
artasparks: nit: Parse have form "Parse parses the root branch"

internal/parser/parser.go
```diff
@@ -0,0 +1,164 @@
...
+// field="PW"
+// value="Player White"
+func (p *Parser) parseFieldValue() (string, string, error) {
+	var tok *scanner.Token
+	var field string
+	var value string
+	// parse a string
+	// TODO: better erroring
+	if tok = p.scanSkipWhitespace(); tok.Type != scanner.String {
+		return "", "", errors.New("Corrupted sgf: 3")
```
artasparks: what does corrupted sgf: 3 mean?

internal/scanner/scanner.go
```diff
@@ -0,0 +1,166 @@
+package scanner
```
artasparks: Generally packages should have the form

// Package scanner does ...

internal/scanner/scanner.go
```diff
@@ -0,0 +1,166 @@
...
+	RightBracket
+	Semicolon
+	Backslash
+	String
+	Illegal
+)
+
+var eof = rune(0)
+
+// isWhiteSpace is a check to see if a rune is whitespace (space, tab, newline, carriage return)
```
artasparks: nit: wrap comments at 80 chars for readability.

internal/scanner/scanner.go
```diff
@@ -0,0 +1,166 @@
...
+	return ch == ' ' || ch == '\t' || ch == '\n' || ch == '\r'
+}
+
+// isSpecial is a check for special characters used in SGF files
+func isSpecial(ch rune) bool {
+	return ch == '(' || ch == ')' || ch == '[' || ch == ']' || ch == ';'
+}
+
+// The scanner will "tokenize" the input into Token structs
+type Token struct {
```
artasparks: Nit: Comment should have form "Token does blah"

Golint should catch these sorts of errors

artasparks (commented):
Still haven't done a full review XD. Couple more comments. I'd recommend running Golint to catch any small errors -- I'll eventually make some presubmits to catch that.

internal/parser/parser.go
```diff
@@ -0,0 +1,164 @@
...
+		return "", "", errors.New("Corrupted sgf: 4")
+	}
+
+	// parse anything until a right bracket
+	for {
+		tok = p.scan()
+		if tok.Type == scanner.Eof {
+			return "", "", errors.New("EOF")
+		}
+		if tok.Type == scanner.RightBracket {
```
artasparks: What happens if you have a right-bracket that's part of a comment? 

[\\]] Should be valid. 

https://github.com/Kashomon/glift-core/blob/master/src/parse/sgf_parsing_test.js has a bunch of edge cases I've collected over the years.

jared-nishikawa:
TIL about golint :)

internal/parser/parser.go
jared-nishikawa: The parser handles `\]`
I'll have a look through the other edge cases

internal/parser/parser.go
jared-nishikawa: Not correctly, I'll tell you that :)

internal/parser/parser.go
jared-nishikawa: If it didn't parse a string, throw an error.  The number is meaningless, just for testing

internal/game/game.go
jared-nishikawa: How about the parser?

artasparks: I'd keep parser public too -- seems like a generally useful tool. Scanner might deserve to be internal, but that could just be private variables within parser

Generally, I feel internal directories should be reserved as a last resort when your private types have grown too unwieldy and you want to share them with other library.

jared-nishikawa: Thanks for clarifying, this is super useful for other projects I have at work

internal/parser/parser.go
jared-nishikawa: Just had a shot at a fix

game/game.go
```diff
@@ -0,0 +1,28 @@
+package game
+
+// Game contains the game tree information
+type Game struct {
+	Root *Node
+}
+
+// NewGame creates a Game struct
+func NewGame() *Game {
```
artasparks: Nit: because the package is Game you can just say New(), because it'll be called as `game.New()`

scanner/scanner.go
```diff
@@ -0,0 +1,168 @@
...
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+)
+
+// constants
+const (
+	EOF = iota
```
artasparks, line 11: nit: You could use a string Consts for these instead of ints.

You also might provide a custome type here for clarity.

type ControlChar int

or similar

Or maybe

type ControlChar rune

scanner/scanner.go
```diff
@@ -0,0 +1,168 @@
...
+	return ch
+}
+
+// Go back one rune
+func (s *Scanner) unread() {
+	s.UnreadRune()
+}
+
+// Scan is the function that tokenizes the input and outputs tokens
+func (s *Scanner) Scan() *Token {
```
artasparks, line 67: nit: I'd recommend passing back the number of chars read so the caller can can keep track of of the column / line # for error messaging. Or maybe # of chars and # of newlines.

scanner/scanner.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	default:
+		return &Token{
+			Type: Illegal,
+			Raw:  "",
+		}
+	}
+}
+
+// scanWhiteSpace consumes as much as whitespace as possible
```
artasparks, line 123: nit: s/scanWhiteSpace/scanWhitespace

scanner/scanner.go
```diff
@@ -0,0 +1,168 @@
...
+var eof = rune(0)
+
+// isWhiteSpace is a check to see if a rune is whitespace
+// (space, tab, newline, carriage return)
+func isWhitespace(ch rune) bool {
+	return ch == ' ' || ch == '\t' || ch == '\n' || ch == '\r'
+}
+
+// isSpecial is a check for special characters used in SGF files
+func isSpecial(ch rune) bool {
```
artasparks, line 32: nit: I might call these ControlChars -- isControlChar

parser/parser_test.go
```diff
@@ -0,0 +1,59 @@
+package parser_test
+
+import (
+	"github.com/otrego/clamshell/parser"
+	"testing"
+)
+
+func TestCreate(t *testing.T) {
```
artasparks, line 8: I'd recommend more extensive tests here especially smaller examples that test specific features. TableTests are a nice way to set this up: https://github.com/golang/go/wiki/TableDrivenTests

artasparks (commented):
Bunch of nits: only thing that I think we should improve before submitting is adding a bit more extensive testing.

artasparks:
I'd like to merge this in so we can start to iterate. Do you have much more you're planning on adding?

jared-nishikawa:
@Kashomon I haven't made all of the changes you've asked for:

- Better testing
- Handling of edge cases
- Custom type for control chars
- Scan() needs to return number of characters scanned

I was thinking I would need at least another week to address all of the edge cases and write better tests.

Option 1: We could merge it in while not in a fully functional state.
Option 2: If your bodoko/sgf_parser is functional, we could just use that.

artasparks:
Let's merge as is and I'll file some issue to follow-up. =)

PR #18, opened 2020-04-26
Initial Continuous Integration with GitHub actions

mhoak:
Uses Github Actions to set up to build and test Otrego.


.github/workflows/go.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14
+      id: go
+    - name: Get dependencies
```
artasparks: We shouldn't need the dep stuff with Go mod. Let's rm.

https://blog.golang.org/using-go-modules

.github/workflows/go.yml
```diff
@@ -0,0 +1,32 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
```
artasparks, line 13: nit: I wonder if ubuntu-slim would work? Might be a little lighter than ubuntu-latest

mhoak: Turns out that only the following are supported:

Windows Server 2019 | windows-latest or windows-2019
Ubuntu 18.04 | ubuntu-latest or ubuntu-18.04
Ubuntu 16.04 | ubuntu-16.04
macOS Catalina 10.15 | macos-latest or macos-10.15


We could run our own runner at some point, but these are ones github supports.

artasparks (approved):

.github/workflows/go.yml
```diff
@@ -0,0 +1,25 @@
...
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14
+      id: go
+    - name: Test
+      run: go test -coverprofile=cover.out ./...
+    - name: Build
```
artasparks, line 24: You shouldn't need to go build -- Go test should do a build, I think.

PR #19, opened 2020-04-26
Move point to core/point.

artasparks:
I'd like to move the core libraries to core/ (game, board, sgf, etc).
This seems like a nice place in the short-term while we use a monorepo.

Also, rm build from github actions. I think it's unnecessary given `go test`


jared-nishikawa (approved):
Looks good to me

PR #29, opened 2020-05-03
Add instructions for installing KataGo for OSX

artasparks:
I went through the process of installing KataGo for OSX. Hopefully this
should save future people some time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,3 +13,51 @@ To run the all the unit tests, run:
 ```shell
 go test ./...
 ```
+
+## Setting up a Dev Environment
+
+This assumes that you have familiarity with and have installed the following:
+
+1.   [Go](https://golang.org/doc/install)
+2.   [Git](https://git-scm.com/book/en/v2)
+
+### Inital OS Setup
+
+*   **OSX**
+    *   We recommend using [Homebrew](https://brew.sh/) to install any packages you may need.
+
+*   **Linux**: TODO. Send us a PR with instructions!
+
+*   **Windows**: TODO. Send us a PR with instructions!
+
+### Installing KataGo
+
+[KataGo](https://github.com/lightvector/KataGo) works with either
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,3 +13,51 @@ To run the all the unit tests, run:
...
+    *   OSX generally comes pre-installed with [OpenCL](https://www.khronos.org/opencl/)
+    *   Run `brew install katago`
+    *   Download the models. Brew comes with a models, which you can get with
+        *   `KATAGO_GTP_CONFIG=$(brew list --verbose katago | grep gtp)`
+        *   `KATAGO_MODEL_PATH= $(brew list --verbose katago | grep .gz | head -1)`
+    *   Tune Katago: `katago benchmark -tune -config $(KATAGO_GTP_CONFIG) -model $(KATAGO_MODEL_PATH)`
+        *   This will output configuration to `$HOME/.katago`
+    *   KataGo should now be operational!
+
+### Using KataGo via UI
```
mhoak, line 56: > Lizzie will also start a Leela Zero process that it will communicate with. You can configure the location of Leela Zero from the config.txt file from the folder you started the java command. If Lizzie is unable to start Leela Zero it will display an error and you can fiddle with the config.txt file until you get all the paths correct.

It doesn't sound like Lizzie will work with Katago at this time.  Given that, I'm not sure that we should add Lizzie as a UI for KataGo at this point. Certainly worth adding as an option. 

[Lizgoban](https://github.com/kaorahi/lizgoban) does appear to work with Katago.

mhoak: Ah - apparently katago it talking to lizzie via GTP. Lizgoban may still be interesting.

artasparks: Yah, I did get Lizzie to work with KataGo via GTP. I'll check out Lizgoban too.

mhoak (approved):
Looks good to me.

PR #33, opened 2020-05-13
Grpc gateway

artasparks:
This adds the proto and code-generation for grpc-gateway. Next,
need to implement the gRPC gateway proxy in the server.


server/api/api.proto
```diff
@@ -19,7 +30,7 @@ message EmptyRequest {}
 
 // FooRequest requests a single Foo
 message FooRequest {
-  string id = 1;
+  string id = 1 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {description: "Foo ID."}];
```
mhoak, line 33: Neat that you can add swagger annotations directly to the proto.

mhoak (commented):
Generally look good. Looking forward to having real API's in these.

mhoak (approved):

PR #34, opened 2020-05-16
Tests

jared-nishikawa:


core/parser/parser_test.go
```diff
@@ -2,58 +2,344 @@ package parser_test
 
 import (
 	"github.com/otrego/clamshell/core/parser"
+	"reflect"
 	"testing"
 )
 
-func TestCreate(t *testing.T) {
-	s := `
+var sgfs = map[string]struct {
```
artasparks: Seems like this should be inlined in the test?

core/parser/parser_test.go
```diff
@@ -2,58 +2,344 @@ package parser_test
...
+	raw string
+	gm  string
```
artasparks: Hmmm. I wonder if we could make this more generic

`type PropMap map[string][]string`

then you could have

`type Path string`

expectedProperties map[Path]PropMap

where a path looks like a treepath from glift: https://github.com/Kashomon/glift/blob/master/deps/glift-core/rules/treepath.js#L40

It's a very useful concept handling SGFs

core/parser/parser_test.go
```diff
@@ -2,58 +2,344 @@ package parser_test
...
+;AW[qc]AB[pd]C[How should White respond?]
+(;W[pc]
+;B[od]C[Correct])
+(;W[qd]
+;B[pe]C[Correct]))`,
+		gm: "1",
+	},
+	"goGameGuruHard": {
+		raw: `
+(;GM[2]FF[4]CA[UTF-8]AP[CGoban:3]ST[2]
```
artasparks: Might consider moving the big blobs to some testdata folder. Here, it seems we're not really testing the properties of the parsed data too much -- i.e., it's not constructed so as to serve the purposes of a unit test.

core/scanner/scanner.go
```diff
@@ -35,8 +37,9 @@ func isSpecial(ch rune) bool {
 
 // Token contains attributes for the Type of token and the raw string
 type Token struct {
-	Type int
+	Type ControlChar
 	Raw  string
+	Num  int
```
artasparks: Is num == len(Raw)? If so, seems like we could remove?

core/parser/parser_test.go
artasparks: Maybe we could validate for just root right now and add treepath support later. Treepath, while very useful, is a rather complex concept.

core/game/game.go
```diff
@@ -13,6 +22,43 @@ func New() *Game {
 	return g
 }
 
+// TreePath returns the node found at a given path
+func (g *Game) TreePath(p Path) (*Node, error) {
```
artasparks, line 26: I think treepath logic should largely live in core/treepath/treepath.go rather than directly in game. It has functionality that is overlapping with but not entirely concurrent with games.

core/game/game.go
```diff
@@ -13,6 +22,43 @@ func New() *Game {
...
+            current = current.Children[treePath[i]]
+        }
+    }
+
+    return current, nil
+}
+
+// parseTreePath currently only handles the simplest possible paths
+// will fix in another branch
+func parseTreePath(p Path) ([]int, error) {
```
artasparks, line 49: Same comment here -- let's move the core/treepath/treepath.go

core/parser/parser_test.go
```diff
@@ -2,58 +2,252 @@ package parser_test
...
+    fname := "base.sgf"
+    expectedProperties := map[game.Path]PropMap{
+        "0": map[string][]string{
+            "GM": []string{"1"},
+        },
+    }
+    testFile(dir, fname, expectedProperties, t)
+}
+
+func TestDescription(t *testing.T) {
```
artasparks, line 66: Can you make these into table tests rather than individual test methods? https://github.com/golang/go/wiki/TableDrivenTests 

That would be more idiomatic

artasparks:
Seems like this got abandoned since it's ~3 mo stale, so I picked this up to unblock the work to get kanalyze off the ground =)

- I've polished some of the testing ideas you introduced here to use treepath
- I formalized treepath (already submitted)
- I reworked the tests to give better error messaging in https://github.com/otrego/clamshell/pull/85

artasparks:
#85 is now submitted.

PR #39, opened 2020-05-19
More refactoring of gRPC

artasparks:


server/grpc/grpc.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/otrego/clamshell/server/echo"
+)
+
+// Options contains options for running gRPC
+type Options struct {
+	// Port to listen on
+	Port int
+}
+
+// Run starts an in-process gRPC gateway and gRPC server.
```
mhoak: The comment is not correct.

mhoak (approved):

server/grpc/grpc.go
artasparks: Done.

PR #41, opened 2020-05-21
Finish basic gRPC / gRPC Gateway service

artasparks:
This allows you to do:

```shell
$ go run main.go
$ curl localhost:10000/v1/echo/foo
{"content":"this is a foo message"}
```

Also, while I'm here, generate and serve swagger.json


mhoak:
I'm reading through this. I don't immediately understand why we'd want to keep two copies of the swagger (api.swagger.go and api.swagger.json) both checked in.

regenerate-proto.sh
```diff
@@ -17,3 +17,21 @@ protoc --grpc-gateway_out=logtostderr=true,paths=source_relative:. \
...
+# Manually go-embed the swagger into a go file.
+#
+# This is quite hacky -- we should use go-bindata or even go generate. For now,
+# it's mega simple. See https://github.com/otrego/clamshell/issues/40
+SWAG_GO="./server/api/api.swagger.go"
+echo 'package api' > "${SWAG_GO}"
+echo '' >> "${SWAG_GO}"
+echo '// Swagger contains embedded swagger data.' >> "${SWAG_GO}"
+echo 'const Swagger = `' >> "${SWAG_GO}"
+cat ./server/api/api.swagger.json | sed 's/`/'"'"'/g' >> "${SWAG_GO}"
```
mhoak: It looks like the api.swagger.json is being generated from the go version and should not be edited. 
It seems like this is what you are pointing to with your comment. 

It would be confusing to know which to update. Maybe it makes sense to move it to a 'generated' folder or possibly rename it to api.generated.swagger.json or something? It also only seems to be used for the swagger endpoint - so maybe we just return the go swagger string? As long as we don't have it duplicated and non-obvious which should be updated.

server/api/api.swagger.json
```diff
@@ -0,0 +1,129 @@
+{
```
mhoak: see comment for regenerate-proto.sh.

regenerate-proto.sh
artasparks: Good point. I've added a line to remove api.swagger.json in the regenerate-proto.sh

I've also added a comment saying that the generated file shouldn't be edited.

server/api/api.swagger.json
artasparks: Ack

jared-nishikawa (approved):

PR #47, opened 2020-05-26
Add a basic test for the echo server

### LINKED ISSUE #43: Add E2E gRPC tests for gRPC Server
With https://github.com/otrego/clamshell/pull/41, I add a basic gRPC service. We should add tests for these basic methods.

Here's some things I found that I haven't look into too deeply

* https://stackoverflow.com/questions/42102496/testing-a-grpc-service. 
* https://godoc.org/google.golang.org/grpc/test/grpc_testing
* https://godoc.org/google.golang.org/grpc/test/bufconn

artasparks:
Fixes https://github.com/otrego/clamshell/issues/43

This adds a simple gRPC test for the echo server.

### DIFF
--- a/server/echo/echo_test.go
+++ b/server/echo/echo_test.go
@@ -1,0 +1,53 @@
+// This is an integration-style test of test.
+package echo
+
+import (
+	"context"
+	"log"
+	"net"
+	"strings"
+	"testing"
+
+	pb "github.com/otrego/clamshell/server/api"
+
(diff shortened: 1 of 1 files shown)


mhoak:
This looks good. 
I think that at some point we'll want to use this as a basis for creating a test-setup for integration style api-tests.

mhoak (approved):

PR #56, opened 2020-06-14
Also run go lint and go vet

artasparks:
Also run go lint and go vet as a github action

Not sure if it will work.


mhoak (approved):
Previously we had build/test as part of one step. 

Should these be separate steps that run in parallel like lint and shadow?

artasparks:
> Previously we had build/test as part of one step.
> Should these be separate steps that run in parallel like lint and shadow?

/shrug. It makes sense to me for them to be serial, since Test depends on Build.

PR #62, opened 2020-06-27
Adds Docker Build.

mhoak:


.github/workflows/go.yml
```diff
@@ -47,30 +91,8 @@ jobs:
         id: plan
         env:
           CONTAINER: api_docker_image=gcr.io/google-samples/hello-app:2.0
-        run: terraform plan -out=plan.tfplan -var="api_docker_image=${CONTAINER}" $GITHUB_WORKSPACE/tf/
+        run: terraform plan -out=plan.tfplan -var="api_docker_image=us.gcr.io/otrego-dev/apiserver:${GITHUB_RUN_NUMBER}" $GITHUB_WORKSPACE/tf/
```
artasparks, line 94: what does GITHUB_RUN_NUMBER mean / correspond to?

.github/workflows/go.yml
```diff
@@ -8,11 +8,11 @@ on:
...
 jobs:
 
-  build:
-    name: Build
+  test:
+    name: Unit Test
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@master
```
artasparks: Is there not a recent release that we can use?

.github/workflows/go.yml
mhoak: "A unique number for each run of a particular workflow in a repository. This number begins at 1 for the workflow's first run, and increments with each new run. This number does not change if you re-run the workflow run."


https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

.github/workflows/go.yml
mhoak: fixed

artasparks (approved):

PR #66, opened 2020-07-05
Fix comments about GOPATH

artasparks:
Fixed some comments about the use of GOPATH. Apparently GOPATH is only used for deps now!

- https://golang.org/doc/gopath_code.html
- https://blog.golang.org/using-go-modules


nelsonjhk (approved):
LGTM

PR #72, opened 2020-07-31
added properties package with validation.

nelsonjhk:
Added a validation method that checks for acceptable SGF tags. Would still need to integrate with parser, but that has a pending PR.


core/parser/prop.go
```diff
@@ -0,0 +1,23 @@
+package prop
```
artasparks: Can you move this to the `game` package?

artasparks: Or, it could have its own package, I suppose.

artasparks: Can you add a test?

https://golang.org/pkg/testing/

core/parser/prop.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+)
+
+
+type Prop struct{
+	prop map["string"][]string
```
artasparks: Could use a method comment. using `golint` will help catch these stylistic things

core/parser/prop.go
```diff
@@ -0,0 +1,23 @@
+package prop
+
+import (
+	"fmt"
+)
+
+
+type Prop struct{
```
artasparks: I was imagining a type for just the key:

`type Prop string`

core/parser/prop.go
```diff
@@ -0,0 +1,23 @@
...
+	prop map["string"][]string
+}
+
+
+func Validate(properties *Properties) (int, error) {
+	var valid = map[string]bool{"AB": true, "AE": true, "AN": true, "AP": true, "AR": true, "AW": true, "B": true, "BL": true, "BM": true, "BR": true, "BT": true, "C": true, "CA": true, "CP": true, "CR": true, "DD": true, "DM": true, "DO": true, "DT": true, "EV": true, "FF": true, "FG": true, "GB": true, "GC": true, "GM": true, "GN": true, "GW": true, "HA": true, "HO": true, "IT": true, "KM": true, "KO": true, "LB": true, "LN": true, "MA": true, "MN": true, "N": true, "OB": true, "ON": true, "OT": true, "OW": true, "PB": true, "PC": true, "PL": true, "PM": true, "PW": true, "RE": true, "RO": true, "RU": true, "SL": true, "SO": true, "SQ": true, "ST": true, "SZ": true, "TB": true, "TE": true, "TM": true, "TR": true, "TW": true, "UC": true, "US": true, "V": true, "VW": true, "W": true, "WL": true, "WR": true, "WT": true,}
```
artasparks: I was imagining defining this at the package level, so the method doesn't have to instantiate it every time.

```
var ValidProperties = map[string]bool{
  ...
}
```

core/parser/prop.go
```diff
@@ -0,0 +1,23 @@
...
+	"fmt"
+)
+
+
+type Prop struct{
+	prop map["string"][]string
+}
+
+
+func Validate(properties *Properties) (int, error) {
```
artasparks: Needs a method comment.

Also, I think you can just return a bool here -- not sure we need (int which seems to be taking the place of bool) or error (the descriptive message doesn't seem terribly helpful)

core/parser/prop.go
artasparks: Oh, I see. This needs to be in its own directory. core/prop/prop.go

PR #73, opened 2020-07-31
Make some fake secrets for testing

artasparks:
Make some fake secrets (private key, cert) for testing. Intent is to
pass in via flags or an env config.


server/secrets/README.md
```diff
@@ -0,0 +1,6 @@
+# Otrego Secrets
+
+This directory contains the secret-logic for Otrego.
+
+
+* `testing/` -- keys / certs for running locally. **never** use in production.
```
dbarnett, line 6: Could you mention:
  a) Sample of how you might use these, and
  b) If they're expected to ever need to be recreated, and how to generate updated equivalents if so?

dbarnett (approved):

artasparks:
Going to close for now, since I'm not actively working on TLS

PR #75, opened 2020-07-31
Remove the use of flags to configure the server, and rely on env vars.

artasparks:
Only a small modification: this changes the server app to use a more
12-factor-y style configuration system that uses environment variables
instead of flags.

This relies extensively on https://github.com/kelseyhightower/envconfig.

For more about 12-factor apps: https://12factor.net/config

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -17,7 +17,8 @@ const (
 	envVarPrefix = "OTREGO"
 )
 
-// Spec is the dynamic configuration spec for the server, which is modified
+// Spec is the dynamic configuration spec for the server, which is modified at
+// runtime via environment variables.
 type Spec struct {
 	// Port for the server, specified with OTREGO_PORT environment variable and
 	// defaulting to 8080 if left unspecified.
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -1,0 +1,35 @@
+// Package contains logic and defaults for the otrego server.
+//
+// Relies on https://github.com/kelseyhightower/envconfig, which is inspired by
+// 12-factor app configuration: https://12factor.net/confi.
+package config
+
+import (
+	"log"
+
+	"github.com/kelseyhightower/envconfig"
+)
+
+const (
+	// envVarPrefix is a prefix to all environment variables.  For example, to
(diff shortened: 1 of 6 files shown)


jared-nishikawa (approved):

server/config/config.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"log"
+
+	"github.com/kelseyhightower/envconfig"
+)
+
+const (
+	// envVarPrefix is a prefix to all environment variables.  For example, to
+	// specify FOO, you will actually need to specify FOO, unless the struct tag
```
dbarnett: I don't understand this "to specify FOO, you will actually need to specify FOO" part. Was it supposed to say "you will actually need to specify OTREGO_FOO"?

server/config/config.go
```diff
@@ -0,0 +1,35 @@
...
+	// `envconfig:"OVERRIDE"` is supplied
+	envVarPrefix = "OTREGO"
+)
+
+// Spec is the dynamic configuration spec for the server, which is modified
```
dbarnett: Did you mean to add more to this sentence after "which is modified"?

server/main.go
```diff
@@ -4,20 +4,20 @@ import (
...
 	"github.com/otrego/clamshell/server/grpc"
 )
 
-var (
-	port = flag.Int("port", 8080, "The server port")
-)
-
 func main() {
-	flag.Set("alsologtostderr", "true")
+	flag.Set("logtostderr", "true")
```
dbarnett, line 12: What does this "logtostderr" vs. "alsologtostderr" change do?

dbarnett (approved):
Nice! *Love* the README content.

server/config/config.go
artasparks: Ah, yes. Fixed

server/config/config.go
artasparks: Oops, thanks.

server/main.go
artasparks: By default, glog logs to a file. alsologtostderr logs to both a file and to stderr.

But you might as well just *not* log to the file and log to stderr. You can always redirect the logs to a file if you want.

PR #77, opened 2020-08-07
Added a validation method that checks for acceptable SGF tags.

### LINKED ISSUE #35: Add better validation for SGF Properties
The SGF spec lives here: https://www.red-bean.com/sgf/. The Go format lives here: https://www.red-bean.com/sgf/go.html

I created something like this here: https://github.com/Kashomon/bodoko/blob/master/core/properties.go

Long time ago. Probably would do it differently now.

nelsonjhk:
Added a prop package that includes a validation method that checks for acceptable SGF tags. Would still need to integrate with parser, but that has a pending PR.


core/prop/prop.go
```diff
@@ -0,0 +1,18 @@
+package prop
```
artasparks, line 1: File needs a package comment.

core/prop/prop.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
```
artasparks: Can removed this

core/prop/prop.go
artasparks, line 1: You'll want to turn go autoformat on: Add let g:go_fmt_autosave = 1 to your .vimrc after you install https://github.com/fatih/vim-go

core/prop/prop.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+
+type Field string
+
+var ValidProperties = map[Field]bool{"AB": true, "AE": true, "AN": true, "AP": true, "AR": true, "AW": true, "B": true, "BL": true, "BM": true, "BR": true, "BT": true, "C": true, "CA": true, "CP": true, "CR": true, "DD": true, "DM": true, "DO": true, "DT": true, "EV": true, "FF": true, "FG": true, "GB": true, "GC": true, "GM": true, "GN": true, "GW": true, "HA": true, "HO": true, "IT": true, "KM": true, "KO": true, "LB": true, "LN": true, "MA": true, "MN": true, "N": true, "OB": true, "ON": true, "OT": true, "OW": true, "PB": true, "PC": true, "PL": true, "PM": true, "PW": true, "RE": true, "RO": true, "RU": true, "SL": true, "SO": true, "SQ": true, "ST": true, "SZ": true, "TB": true, "TE": true, "TM": true, "TR": true, "TW": true, "UC": true, "US": true, "V": true, "VW": true, "W": true, "WL": true, "WR": true, "WT": true,}
```
artasparks: nit: for readability, maybe break into multiple lines

core/prop/prop.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Validate returns true if Prop is an accepted SGF property.
+func Validate(field Field) (bool) {
+	if _, ok := ValidProperties[field]; ok {
```
artasparks: a bit simpler:

_, ok := ValidProperties[field]
return ok

core/prop/prop_test.go
```diff
@@ -0,0 +1,25 @@
+package prop_test
```
artasparks, line 1: just `package prop`

artasparks, line 1: needs to be formatted

core/prop/prop_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	var valid prop.Field = "TB"
+	var invalid prop.Field = "TQB"
+
+	test := prop.Validate(valid)
+	if test != true{
+		t.Errorf("Valid SGF field returned %v; want true", test)
+	}
+
+	test = prop.Validate(invalid)
```
artasparks: probably these should be separate tests.

core/prop/prop.go
artasparks: Looks like this still needs to happen

core/prop/prop_test.go
artasparks: I think this is duplicated now? So just remove

core/prop/prop.go
nelsonjhk: Comment is below deleted lines (new line 3)

core/prop/prop_test.go
nelsonjhk: Ah, forgot to clean up

core/prop/prop.go
artasparks: Ah, comment should be above the package:

```go
// Package prop does a thing
package prop
```

But at this point, it's optional, so easy to clean up later =).

artasparks (approved):

PR #79, opened 2020-09-01
Added Point Translation Methods to Package Point

KiloLimaFoxtrot:
Point translation methods have been added to the Package point. The methods translate from Int points to SGF points, and vice versa. -Kilo


core/point/point.go
```diff
@@ -21,3 +25,34 @@ func (p *Point) X() int64 { return p.x }
 
 // Y returns the y-value.
 func (p *Point) Y() int64 { return p.y }
+
+// Kilo Foxtrot addition below
```
artasparks: Rm these lines

core/point/point.go
```diff
@@ -21,3 +25,34 @@ func (p *Point) X() int64 { return p.x }
...
+// Three point translation methods, one is an alternate
+
+// 1. '*Point' immutable int point, to SGF(String of two letters) Point
+func translateToSGFPt01(pt *Point) string {
```
artasparks: The naming is a little wordy. How about just ToSGF()

Also to note: lowerCaseNames are private, UpperCaseNames are public.

artasparks: How about we make this a method on the Point object

func (pt *point) ToSGF() string {

}

core/point/point.go
```diff
@@ -21,3 +25,34 @@ func (p *Point) X() int64 { return p.x }
...
+	sgfPtX := string(rune((pt.X()) + 97))
+	sgfPtY := string(rune((pt.Y()) + 97))
+	// two lower case letters
+	// return sgfPtX + sgfPtY
+	return strings.ToLower(sgfPtX) + strings.ToLower(sgfPtY)
```
artasparks: Upper case and lower case are both valid in coordinates, technically. So, I wouldn't artificially make it lower-case.

core/point/point.go
```diff
@@ -21,3 +25,34 @@ func (p *Point) X() int64 { return p.x }
...
+}
+
+// Alternate, in case the incoming point is not immutable
+// 2. 'Point' mutable int point, to SGF(String of two letters) Point
+func translateToSGFPt02(pt Point) string {
```
artasparks: This looks identical to the previous. what am I missing?

Probably just remove

core/point/point.go
```diff
@@ -21,3 +25,34 @@ func (p *Point) X() int64 { return p.x }
 
 // Y returns the y-value.
 func (p *Point) Y() int64 { return p.y }
+
+// Kilo Foxtrot addition below
+// Three point translation methods, one is an alternate
+
+// 1. '*Point' immutable int point, to SGF(String of two letters) Point
+func translateToSGFPt01(pt *Point) string {
+	sgfPtX := string(rune((pt.X()) + 97))
```
artasparks: instead of subtracting 97, how about storing as a const?

const aValue = int('a')

Then here you can just add pt.X() + aValue

core/point/point.go
```diff
@@ -21,3 +25,34 @@ func (p *Point) X() int64 { return p.x }
...
+func translateToSGFPt02(pt Point) string {
+	sgfPtX := string(rune((pt.X()) + 97))
+	sgfPtY := string(rune((pt.Y()) + 97))
+	// two lower case letters
+	// return sgfPtX + sgfPtY
+	return strings.ToLower(sgfPtX) + strings.ToLower(sgfPtY)
+}
+
+// 3. 'p' SGF(two letter string) Point, to '*Point' immutable Int Point
+func translateToINTPt01(sgfPt string) *Point {
```
artasparks: Again, you can make clearer with naming:

func NewFromSGF(pt string) *Point {

}

artasparks (commented):
Can you add some unit tests? See 

- https://golang.org/pkg/testing/
- https://blog.alexellis.io/golang-writing-unit-tests/

core/point/point.go
artasparks: Oh, I see. Point is a non-pointer type. I'd just remove this and rely on the Pointer version.

core/point/pointTranslation_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"testing"
+)
+
+// Two sample correct results
+// New(8, 20), SGF string "iu"
+// New(13, 5), SGF string "nf"
+
+func TestTranslate(t *testing.T) {
+	
+	fmt.Println()
```
artasparks, line 14: rm the extra fmt.Printlns. Looks like for debugging

core/point/pointTranslation_test.go
```diff
@@ -0,0 +1,91 @@
...
+	fmt.Println("[*** Go Game Point Translation Tests ***]")
+	
+	// Test Point Struct
+	type PointStrct struct {
```
artasparks, line 18: This is a pretty wordy test construction and quite hard to read. I would recommend using Table-based tests.

https://github.com/golang/go/wiki/TableDrivenTests

core/point/point.go
```diff
@@ -21,3 +23,21 @@ func (p *Point) X() int64 { return p.x }
...
+	return sgfX + sgfY
+}
+
+// 3. 'p' SGF(two letter string) Point, to '*Point' immutable Int Point
+func NewFromSGF(sgfPt string) *Point {
+	sgfX := sgfPt[0]
+	sgfY := sgfPt[1]
+	x := int64(sgfX) - aValue
+	y := int64(sgfY) - aValue
+	return New(int64(x), int64(y))
```
artasparks, line 42: I think you can condense this  a bit. Ex: `x := int64(sgfPt[0]) - aValue`

core/point/point.go
```diff
@@ -21,3 +23,21 @@ func (p *Point) X() int64 { return p.x }
...
+// 1. '*Point' immutable int point, to SGF(String of two letters) point
+func (pt *Point) ToSGF() string {
+	sgfX := string(rune((pt.X()) + aValue))
+	sgfY := string(rune((pt.Y()) + aValue))
+	// two lower case letters
+	return sgfX + sgfY
+}
+
+// 3. 'p' SGF(two letter string) Point, to '*Point' immutable Int Point
+func NewFromSGF(sgfPt string) *Point {
```
artasparks, line 37: Try running golint: https://github.com/golang/lint

Typically methods have the following pattern for comments:

// ToSGF converts a SGF point to a Point. Example: ab => {0, 1}

Also, you should be clear about whether the returned point is 0-indexed or 1-indexed. I think it probably makes sense to make it 0-indexed.

core/point/point.go
```diff
@@ -21,3 +23,21 @@ func (p *Point) X() int64 { return p.x }
 
 // Y returns the y-value.
 func (p *Point) Y() int64 { return p.y }
+
+// Two Go-game board position/point translation methods
+// 1. '*Point' immutable int point, to SGF(String of two letters) point
+func (pt *Point) ToSGF() string {
+	sgfX := string(rune((pt.X()) + aValue))
+	sgfY := string(rune((pt.Y()) + aValue))
+	// two lower case letters
```
artasparks, line 32: rm comment. Not necessarily true.

core/point/point.go
artasparks, line 37: What if the input is malformed? I.e., if sgfPt == "abc" or "a" 

Right now what will happen? What should happen?

core/point/pointTranslation_test.go
```diff
@@ -0,0 +1,91 @@
+package point
```
artasparks, line 1: You can just include these tests in point_test.go

artasparks:
Moved to PR #84

PR #80, opened 2020-09-05
Add a katago analysis wrapper

artasparks:
Not much here, mostly just a wrapper, but it seems like it's a good checkpoint right now before adding JSON analysis.

### DIFF
--- a/core/katago/katago.go
+++ b/core/katago/katago.go
@@ -1,12 +1,6 @@
 // Package katago provides wrappers for analyzing games with katago.
 package katago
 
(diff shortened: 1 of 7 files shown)


nelsonjhk (approved):

nelsonjhk (approved):
lgtm; the types in query.go might be something to standardize across Clamshell?

artasparks:
Yeah, I think that's possible. Especially, something like 'rules' and 'moves.' I'll look into it for a follow-up =).

PR #81, opened 2020-09-08
Add color / move helpers

artasparks:
Follow up from previous PR (adding some analysis tools): Add a color and move helpers.

### DIFF
--- a/core/color/color_test.go
+++ b/core/color/color_test.go
@@ -1,0 +1,43 @@
+package color
+
+import (
+	"testing"
+)
+
+func TestOpposite(t *testing.T) {
+	testCases := []struct {
+		desc string
+		in   string
+		want string
+	}{
+		{
+			desc: "white=>black",
+			in:   "W",
+			want: "B",
+		},
+		{
+			desc: "black=>white",
+			in:   "B",
+			want: "W",
+		},
+		{
+			desc: "empty=>empty",
+			in:   "",
+			want: "",
+		},
+		{
+			desc: "any=>any",
+			in:   "any",
+			want: "any",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			out := Color(tc.in).Opposite()
+			if out != Color(tc.want) {
+				t.Errorf("%q.Opposite()=%q, but wanted %q", tc.in, out, tc.want)
+			}
+		})
+	}
+}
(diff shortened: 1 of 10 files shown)


core/color/color_test.go
```diff
@@ -0,0 +1,43 @@
+package color
+
+import (
+	"testing"
+)
+
```
KiloLimaFoxtrot, line 6: The test table, with its slice array of test input and output values, makes how this test method works clear. Further, the design of this test method is very good, being as simple as possible, but not simpler, and able to test critical Opposite() color translation functionality

core/color/color_test.go
```diff
@@ -0,0 +1,43 @@
...
+			desc: "black=>white",
+			in:   "B",
+			want: "W",
+		},
+		{
+			desc: "empty=>empty",
+			in:   "",
+			want: "",
+		},
+		{
```
KiloLimaFoxtrot, line 28: Good to distiinguish between the test conditions of an empty string and a random string, even though both return the same general result; ie the same value as was entered, because the two inputs obviously have different data types.

core/color/color_test.go
```diff
@@ -0,0 +1,43 @@
...
+			desc: "any=>any",
+			in:   "any",
+			want: "any",
+		},
+	}
+
```
KiloLimaFoxtrot, line 34: The for loop here below effectively iterates through the test struct input and output values and both outputs the test results and a simple string description of each iteration's input and output struct testCases' slice arrays values. The string description facilitates following the logic of this test method, and identifying where failures may occur. Entering bad data on the color.go side causes this test to fail consitently, triggering the 'if out != Color(tc.want) {...}' conditional below, demonstrating that this test is effective.

core/color/color_test.go
```diff
@@ -0,0 +1,43 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			out := Color(tc.in).Opposite()
+			if out != Color(tc.want) {
+				t.Errorf("%q.Opposite()=%q, but wanted %q", tc.in, out, tc.want)
+			}
+		})
+	}
+}
```
KiloLimaFoxtrot, line 43: The simple design, effective testing with test tables; which allow for the iteration over a complete spectrum of critical test parameters, makes this test method a positive improvement to the Otrego code base. 

I will use the color.go and color_test.go as a model for the point translation and test methods.

KiloLimaFoxtrot (approved):
Good job. And thanks, this makes several things clearer to me about the relationship between core methods and their test methods.

PR #82, opened 2020-09-10
Add a treepath implementation

artasparks:
This adds a clean-slate implementation of the treepath from gliftgo: https://github.com/Kashomon/glift-core/blob/master/src/rules/treepath.js

Treepath is one of the more useful things I created for Glift, so makes perfect sense to port-over.

Note: I drop the plus syntax (4+ or 0.2+) because it was never really used. The 2:3 was more versatile and useful, ultimately.

### DIFF
--- a/core/treepath/treepath.go
+++ b/core/treepath/treepath.go
@@ -11,6 +11,8 @@ import (
 	"strconv"
 	"strings"
 	"unicode"
+
+	"github.com/otrego/clamshell/core/game"
 )
 
 // A Treepath is a list of variations that says how to travel through a tree of
@@ -136,12 +138,12 @@ func ParseInitialPath(initPos string) (Treepath, error) {
 type parseState int
 
 const (
-	// variation means we are looking for numbers for the variation number. Can
+	// variationState means we are looking for numbers for the variation number. Can
 	// either transition to either separator (flush) or repeat (flush and then
 	// repeat N times
-	variation parseState = iota
+	variationState parseState = iota
 	// repeat means we repeat the previous variation 'n' times.
-	repeat
+	repeatState
 )
 
 // ParseFragment parses a treepath fragment.
@@ -154,7 +156,7 @@ const (
 // 53 = [53]
 func ParseFragment(path string) (Treepath, error) {
 	out := Treepath{}
-	curState := variation
+	curState := variationState
 	buf := &strings.Builder{}
 	prevChar := 0
 
@@ -170,8 +172,24 @@ func ParseFragment(path string) (Treepath, error) {
 		return n, nil
 	}
 
-	// Use a sentinal newline character
+	// Use a sentinal newline character so that it's clear when to flush the final
+	// charactrs.
 	path += "\n"
+
+	//
+	// The rough approach is to have a simple parser modeled as a 2-state DFA.
+	//
+	// VARIATION: means we're parsing a normal variation number. We flush the
+	// buffer data when we hit a '.' rune.
+	//
+	// REPEAT: means we have seen a ':' instead of a '.', so we're going to repeat
+	// the previous number once we parse the next number.
+	//
+	//  --'.'--- <------------
+	//  |      |             |
+	//  V      ^             ^
+	// VARIATION --':'-->  REPEAT
+	//
 	for idx, c := range path {
 		if unicode.IsDigit(c) {
 			buf.WriteRune(c)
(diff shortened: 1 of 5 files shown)


core/treepath/treepath.go
```diff
@@ -0,0 +1,257 @@
...
+// A Treepath is a list of variations that says how to travel through a tree of
+// moves. And has two forms a list version and astring version. First, the
+// list-version:
+//
+//    [0,1,0]
+//
+// Means we will first take the 0th variation, then we will take the 1ist
+// variation, and lastly we will take the 0th variation again. For
+// convenience, the treepath can also be specified by a string, which is where
+// the fun begins. At it's simpliest,
```
dbarnett: Typo: "it's simpliest" -> "its simplest"

core/treepath/treepath.go
```diff
@@ -0,0 +1,257 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+	"unicode"
+
+	"github.com/otrego/clamshell/core/game"
+)
+
+// A Treepath is a list of variations that says how to travel through a tree of
+// moves. And has two forms a list version and astring version. First, the
```
dbarnett: Typo? "astring" -> "a string"

core/treepath/treepath.go
```diff
@@ -0,0 +1,257 @@
...
+//
+//    [0,0,0] becomes 0.0.0
+//
+// but there are a couple different string short-hands that make using treepaths
+// a little easier
+//
+//    0.1:2   Take the 0th variation, then repeat taking the 1st varation twice
+//
+// There are two types of treepaths discussed below -- A *treepath fragment*
+// (which is what we have been describing) and an *initial treepath*.
```
dbarnett: Can this description give a sense of where each would be used?

It seems like the advantage of initial treepaths is that they're slightly shorter, but IIUC only saving 2 characters while adding some conceptual overhead: `0:53` vs. `53`. Is there more to it than that?

artasparks: Awesome. Love it. This is a great simplification.

core/treepath/treepath.go
artasparks: Done

core/treepath/treepath.go
artasparks: Done

core/treepath/treepath.go
```diff
@@ -0,0 +1,173 @@
...
+// * Parse("1.2:1.0.2:3") => [1,2,0,2,2,2]
+func Parse(path string) (Treepath, error) {
+	out := Treepath{}
+	curState := variationState
+	buf := &strings.Builder{}
+	prevChar := 0
+
+	convertBuffer := func(idx int) (int, error) {
+		if buf.Len() == 0 {
+			return 0, fmt.Errorf("error parsing fragment %q at index %d: separator '.' must be proceeded by digit", path, idx)
```
dbarnett: For the error message here, it could be not a `.` but a `:` or EOF, right? Think it's worth the complexity to have correct error messages?

core/treepath/treepath.go
```diff
@@ -0,0 +1,173 @@
...
+// Apply applies a treepath to a node, moving down a tree of moves. This
+// takes the variation listed in the treepath until either:
+//
+// 1. There are no more variation numbers in the treepath.
+// 2. There is not a child with the given variation number.
+func (tp Treepath) Apply(n *game.Node) *game.Node {
+	curNode := n
+	for _, v := range tp {
+		if v < len(curNode.Children) {
+			fmt.Printf("var %v\n", v)
```
dbarnett: Is this output intentional, basically info logging going to stdout for now?

core/treepath/treepath.go
dbarnett: Should it not say there are two types anymore?

Still useful to note the idiom for skipping a series of initial moves if it's especially common, it's just not a standalone "type" anymore.

dbarnett (approved):
:+1:

core/treepath/treepath.go
artasparks: Done

core/treepath/treepath.go
artasparks: Done

core/treepath/treepath.go
artasparks: Done

PR #84, opened 2020-09-12
Updated point.go and point_test.go

KiloLimaFoxtrot:
Updated point.go and point_test.go. Included test tables, some basic translation function error handling and changed the translation method from ASCII to a slice of chars/bytes a to z, lower and upper case.

### DIFF
--- a/core/point/point.go
+++ b/core/point/point.go
@@ -1,6 +1,10 @@
 // Package point is a basic package for points.
 package point
 
+import (
+	"fmt"
+)
+
 // Point is a basic point. Although simple, the member variables are kept
 // private to ensure that Point remains immutable.
 type Point struct {
@@ -8,9 +12,9 @@ type Point struct {
 	y int64
 }
 
-// PointToSgfMap is a translation reference between int64 Point
+// pointToSgfMap is a translation reference between int64 Point
 // and string SGF-Point (rune) values
-var PointToSgfMap = map[int64]rune{
+var pointToSgfMap = map[int64]rune{
 	0: 'a', 1: 'b', 2: 'c', 3: 'd', 4: 'e', 5: 'f', 6: 'g',
 	7: 'h', 8: 'i', 9: 'j', 10: 'k', 11: 'l', 12: 'm', 13: 'n',
 	14: 'o', 15: 'p', 16: 'q', 17: 'r', 18: 's', 19: 't', 20: 'u',
@@ -21,9 +25,9 @@ var PointToSgfMap = map[int64]rune{
 	49: 'X', 50: 'Y', 51: 'Z',
 }
 
-// SgfToPointMap is a translation reference between string SGF-Point
+// sgfToPointMap is a translation reference between string SGF-Point
 // (rune) values and int64 Point values
-var SgfToPointMap = map[rune]int64{
+var sgfToPointMap = map[rune]int64{
 	'a': 0, 'b': 1, 'c': 2, 'd': 3, 'e': 4, 'f': 5, 'g': 6, 'h': 7,
 	'i': 8, 'j': 9, 'k': 10, 'l': 11, 'm': 12, 'n': 13, 'o': 14,
 	'p': 15, 'q': 16, 'r': 17, 's': 18, 't': 19, 'u': 20, 'v': 21,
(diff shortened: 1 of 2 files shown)


core/point/point.go
```diff
@@ -1,13 +1,25 @@
...
 // Point is a basic point. Although simple, the member variables are kept
 // private to ensure that Point remains immutable.
 type Point struct {
 	x int64
 	y int64
 }
 
+// PointSGFSlce is a slice/Array translation reference between integer
+// -Point (index position) and string SGF-Point (byte/char) values
+var PointSGFSlce = []byte{'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h',
```
artasparks: How about you this a map so you don't have to scan?

ex: map[rune]int{
'a': 1
'b': 2
...
}

core/point/point.go
```diff
@@ -17,7 +29,38 @@ func New(x, y int64) *Point {
...
+// ToSGF converts a pointer-type (immutable) *Point
+// to an SGF Point (two letter string). The returned value is 0-indexed.
+func (pt *Point) ToSGF() string {
+	sgfOut := ""
+	if (pt != nil) && (pt.X() <= 51) && (pt.Y() <= 51) {
+		sgfX := string(PointSGFSlce[pt.X()])
+		sgfY := string(PointSGFSlce[pt.Y()])
+		sgfOut = sgfX + sgfY
+	} else {
+		panic("Error: *Point entries must not be nil for either" +
```
artasparks: Don't panic -- just return an error as a value. Panic is for exceptional cases and programming errors -- stuff that should cause the program to crash.

- https://blog.golang.org/error-handling-and-go

core/point/point_test.go
```diff
@@ -13,3 +13,117 @@ func TestCreate(t *testing.T) {
 		t.Errorf("p.Y()=%v, expected %v", gotY, 2)
 	}
 }
+
+func TestPointSGFTranslation(t *testing.T) {
+	type Point struct {
```
artasparks: No reason to define this here -- just use the one from point.go

core/point/point_test.go
```diff
@@ -13,3 +13,117 @@ func TestCreate(t *testing.T) {
...
+			in: Point{
+				x: 40,
+				y: 51,
+			},
+			want: "OZ",
+		},
+	}
+
+	for _, tc := range testToSGFCases {
+		t.Run(tc.desc, func(t01 *testing.T) {
```
artasparks: t01 should just be t. It's ok to shadow t in this context.

core/point/point_test.go
```diff
@@ -13,3 +13,117 @@ func TestCreate(t *testing.T) {
...
+			toSGFOut := New(tc.in.x, tc.in.y).ToSGF()
```
artasparks: The copying here from your custom point is unnecessary -- just use the point from point.go

core/point/point_test.go
```diff
@@ -13,3 +13,117 @@ func TestCreate(t *testing.T) {
...
+			if t01 == nil || toSGFOut != tc.want {
```
artasparks: why would t01 be nil? that should be impossible -- not something you need to check for. It's ok to let such a nil pointer exception/panic (NPE) happen here.

core/point/point_test.go
```diff
@@ -13,3 +13,117 @@ func TestCreate(t *testing.T) {
...
+				t01.Errorf("%q.ToSGF() = %q, but wanted %q", tc.in,
+					toSGFOut, tc.want)
+			}
+		})
+	}
+
+	// Second test translation from SGF-string-point to integer-point
+	testToPointCases := []struct {
```
artasparks, line 70: I would make this a different test method TestConvertSGFToPoint

core/point/point_test.go
```diff
@@ -13,3 +13,117 @@ func TestCreate(t *testing.T) {
...
+			in:   "OZ",
+			want: &Point{
+				x: 40,
+				y: 51,
+			},
+		},
+	}
+
+	for _, tc := range testToPointCases {
+		t.Run(tc.desc, func(t01 *testing.T) {
```
artasparks: again, you can shadow `t` here -- just use `t` instead of `t01`. That's one of the intents of this pattern.

core/point/point_test.go
```diff
@@ -13,3 +13,117 @@ func TestCreate(t *testing.T) {
...
+			toPointOut := NewFromSGF(tc.in)
+			// include the point.go *Point type X Y getters below
+			pointX := toPointOut.X()
+			pointY := toPointOut.Y()
+			if t01 == nil || pointX != tc.want.x {
```
artasparks: rm check that t01 is non-nil -- that should never happen.

artasparks (commented):
Awesome! This now looks like go-code! Just have some relatively minor cleanups now.

core/point/point_test.go
KiloLimaFoxtrot: Not clear on what you are refereing to here. Thanks.

core/point/point.go
KiloLimaFoxtrot: That is a good idea. A map would be more efficient. However, I'm not sure in GoLang how to reverse map, ie retrieve a key from a value, as when translating from Point to SGF and vice versa. Hence, I made two maps. One has the rune/char values as keys, and one has the int64 values as keys. Also, I left the whole thing 0 indexed as we had previously discussed, and because non-zero based seemed to have some issues and require pushing the translations +1 or -1 to make them work. Let me know if I am missing something. I will be submitting a pull request shortly.

KiloLimaFoxtrot: Actually, my update simply updated this pull request. The changes can be seen in the files changed section. Thanks!

core/point/point.go
```diff
@@ -8,6 +8,32 @@ type Point struct {
 	y int64
 }
 
+// PointToSgfMap is a translation reference between int64 Point
+// and string SGF-Point (rune) values
+var PointToSgfMap = map[int64]rune{
```
artasparks: should be private problably.

core/point/point.go
```diff
@@ -8,6 +8,32 @@ type Point struct {
...
+	21: 'v', 22: 'w', 23: 'x', 24: 'y', 25: 'z', 26: 'A', 27: 'B',
+	28: 'C', 29: 'D', 30: 'E', 31: 'F', 32: 'G', 33: 'H', 34: 'I',
+	35: 'J', 36: 'K', 37: 'L', 38: 'M', 39: 'N', 40: 'O', 41: 'P',
+	42: 'Q', 43: 'R', 44: 'S', 45: 'T', 46: 'U', 47: 'V', 48: 'W',
+	49: 'X', 50: 'Y', 51: 'Z',
+}
+
+// SgfToPointMap is a translation reference between string SGF-Point
+// (rune) values and int64 Point values
+var SgfToPointMap = map[rune]int64{
```
artasparks: nit (not required to change): I'd recommend making a simple, anonymous function that runs and inverts the previous map


```go
var sgfToPoint = func(m map[int64]rune) map[rune]int64 {
  o := make(map[rune]int64)
  for key, val := range m {
   o[val] = key
  }
  return o
}(pointToSGF)
```

core/point/point.go
```diff
@@ -17,7 +43,38 @@ func New(x, y int64) *Point {
...
+// to an SGF Point (two letter string).
+// The returned value is 0-indexed.
+func (pt *Point) ToSGF() string {
+	sgfOut := ""
+	if (pt.X() <= 51) && (pt.Y() <= 51) {
+		sgfX := string(PointToSgfMap[pt.X()])
+		sgfY := string(PointToSgfMap[pt.Y()])
+		sgfOut = sgfX + sgfY
+	} else {
+		sgfOut = "--"
```
artasparks: I'd recommend returning an actual error via fmt.Errorf.

That would make the signature 

```go
func (pt *Point) ToSGF() (string, error) {
...
}
```

I can follow up with this change if you'd like

core/point/point.go
```diff
@@ -17,7 +43,38 @@ func New(x, y int64) *Point {
...
+	}
+	return sgfOut
+}
+
+// NewFromSGF converts an SGF point (two letter string)
+// to a pointer-type (immutable) *Point.
+func NewFromSGF(sgfPt string) *Point {
```
artasparks: Same comment here, I would return an error on bad data.

I can take this or you can.

artasparks (approved):
Approved. I think the general shape of the code is quite good. I can clean up the remaining bits now, or you can. Up to you.

KiloLimaFoxtrot:
Refined point.go and point_test.go with translation fmt.Errof parameters. This update foes not include the Nit suggestion, which can be done if recommended.

artasparks (approved):
Looks good

Issue #83, opened 2020-09-11
Parsing adds extra node before root


artasparks:
It looks like the parser produces an extra node at the root node. Normally, I would expect the root node to be all the properties that set up the game (AW, BW, GM[1], etc.).

But right now it looks like

Game => Root (Empty) => Actual root

artasparks:
Found in https://github.com/otrego/clamshell/pull/82

artasparks:
Fixed in #85

Issue #21, opened 2020-04-26
Improvements for Parser


artasparks:
It would be nice to improve the parser/scanner to be more robust / give better error signals. In particular, we should have:

* Better testing
* Handling of edge cases
* Custom type for control chars
* Scan() needs to return number of characters scanned

jared-nishikawa:
The custom type and adjustments to scan() were straightforward.  I just made a PR with another iteration of testing and edge cases.  Not necessarily complete, but at least a tidier framework for testing.  Interested in your thoughts regarding the structure

artasparks:
Done with #85

PR #85, opened 2020-09-13
Rework the parser

artasparks:
This PR completely reworks the parser, based on https://github.com/Kashomon/bodoko/blob/master/core/sgf_parser.go

- Use a DFA (state-based) single-pass parser rather than a scanner with backtracking. No backtracking is needed now and I find it much easier to reason about.
- Add line/col/index tracking for better error messaging
- Add extensive tests, relying on treepath
- Re-enable treepath test for Apply
- Add a special-case treepath '.' to indicate an empty treepath -- useful for grabbing the root node.

Fixes: 
- https://github.com/otrego/clamshell/issues/83
- https://github.com/otrego/clamshell/issues/21


core/sgf/parser.go
```diff
@@ -0,0 +1,422 @@
...
+
+		switch sd.curstate {
+		case beginningState:
+			err = handleBeginning(sd, pbuf, g)
+			if err != nil {
+				return nil, err
+			}
+
+		case betweenState:
+			err = handleBetween(sd, pbuf)
```
artasparks:
```
if err := handleBetween(sd, pbuf); err != nil {
  return err
}
```

core/game/node.go
```diff
@@ -0,0 +1,32 @@
...
+	// Move indicates a move in the game.
+	Move *Move
+
+	// Properties contain all the raw/unprocessed properties
+	Properties map[string][]string
+
+	// Children of this position.
+	Children []*Node
+
+	//
```
artasparks: Fix this comment

artasparks: Fixed, also fixed logic to add parent.

core/sgf/parser.go
artasparks: Done.

core/sgf/parser.go
```diff
@@ -0,0 +1,399 @@
...
+//
+//     propData => propData
+//     propData => between
+func handlePropData(stateData *stateData, pbuf *propBuffer) error {
+	if stateData.curchar == rbrace &&
+		// C[foo 1[k\] bar]
+		//           ^
+		// There was a backslash used for escaping a r-brace.
+		stateData.holdChar == backslash {
+		stateData.holdChar = rune(0)
```
mhoak, line 337: Is the formatting off in these if-statements? Part of the if statement conditional is in line with the if branch. That seems wrong.

mhoak (approved):

core/sgf/parser.go
artasparks: It's technically correct, but confusing. Fixed

PR #90, opened 2020-09-15
add String() method to Point

### LINKED ISSUE #89: Add String method to Point
We should add `String()` to Point, which will be useful during tests (since "%v" will use it).

ilmanzo:
fix #89

### DIFF
--- a/core/point/point.go
+++ b/core/point/point.go
@@ -74,6 +74,11 @@ func (pt *Point) ToSGF() (string, error) {
 	return sgfOut, err01
 }
 
+// String() method to represent and print a Point, useful for debugging and test purposes
+func (pt Point) String() string {
+	return fmt.Sprintf("Point(%d,%d)", pt.x, pt.y)
+}
+
 // NewFromSGF converts an SGF point (two letter string)
 // to a pointer-type (immutable) *Point.
 func NewFromSGF(sgfPt string) (*Point, error) {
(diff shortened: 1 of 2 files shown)


core/point/point.go
```diff
@@ -74,6 +74,11 @@ func (pt *Point) ToSGF() (string, error) {
 	return sgfOut, err01
 }
 
+// String() method to represent and print a Point, useful for debugging and test purposes
+func (pt Point) String() string {
+	return fmt.Sprintf("Point(%d,%d)", pt.x, pt.y)
```
artasparks: optional: in go, typically you would just put the data `{x,y}` rather than the type. I think that's a little more compact.

core/point/point_test.go
```diff
@@ -14,6 +17,15 @@ func TestCreate(t *testing.T) {
 	}
 }
 
+func TestPointToString(t *testing.T) {
+	p := New(2, 3)
+	result := fmt.Sprintf("%v", p)
```
artasparks: optional: you could inline p here: `fmt.Sprintf("%v", New(2, 3))`

artasparks (commented):
Just some optional comments. 

I'm curious how you found this repository given its low numbers of stars and the fact that we haven't advertised it =)

ilmanzo:
> Just some optional comments.
> 
> I'm curious how you found this repository given its low numbers of stars and the fact that we haven't advertised it =)

Hi, I am both a novice go (game) player and a beginner go (language) developer, so I like your project :) I found it on https://issuehub.io searching by tag ...

artasparks:
Cool, thanks for the PR!

artasparks (approved):

PR #94, opened 2020-09-26
Cleaned Up Point Class

KiloLimaFoxtrot:
Added the reverse Map function that creates a reference for the reverse translation. 

Added the (0 <= pt.X()) and (0 <= pt.Y()) input error checks/parameters to the ToSGF() function. 

Removed irrelevant commented code and other artifacts from previous versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ tmp
 tf/.terraform
 tf/*.plan
 tf/*.state
+
+# katago analysis logs
+analysis.log
--- a/core/katago/katago.go
+++ b/core/katago/katago.go
@@ -1,11 +1,51 @@
 // Package katago provides wrappers for analyzing games with katago.
 package katago
 
-// Analyzer is a katago-analyzer
+import (
+	"io"
+	"os/exec"
+	"strconv"
+
+	"github.com/golang/glog"
+)
+
+// Analyzer is a katago-analyzer.
 type Analyzer struct {
 	// Model is the path to the model file.
 	Model string
 
 	// Config is the path to the config file.
 	Config string
+
+	// Number of analysis threads. Defaults to 16 if not specified.
+	AnalysisThreads int
+}
+
+// Cmd creates the Katago analysis command.
+func (an *Analyzer) Cmd() *exec.Cmd {
+	threads := an.AnalysisThreads
+	if threads == 0 {
+		threads = 8
+	}
+	return exec.Command("katago", "analysis", "-model", an.Model, "-config", an.Config, "-analysis-threads", strconv.Itoa(threads))
+}
+
+// Start the katago analyzer.
+func (an *Analyzer) Start() (io.WriteCloser, io.ReadCloser, error) {
+	glog.Info("Starting katago analyzer")
+
+	cmd := an.Cmd()
+	stdinPipe, err := cmd.StdinPipe()
+	if err != nil {
+		return nil, nil, err
+	}
+	stdoutPipe, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, nil, err
+	}
+	err = cmd.Start()
+	if err != nil {
+		return nil, nil, err
+	}
+	return stdinPipe, stdoutPipe, nil
 }
--- a/core/katago/query.go
+++ b/core/katago/query.go
@@ -2,8 +2,12 @@ package katago
 
 import (
 	"encoding/json"
+	"fmt"
+	"strconv"
 
 	"github.com/google/uuid"
+	"github.com/otrego/clamshell/core/game"
+	"github.com/otrego/clamshell/core/point"
 )
 
 // Query is a katago analysis query. Here we use pointer values for primitives
@@ -13,7 +17,10 @@ type Query struct {
 	ID string `json:"id"`
 
 	// InitialStones are the initial stones on the board.
-	InitialStones []Move `json:"initialStones"`
+	InitialStones []Move `json:"initialStones,omitempty"`
+
+	// InitialPlayer is the initial player
+	InitialPlayer string `json:"initialPlayer,omitempty"`
 
 	// Moves are the moves played in the game.
 	Moves []Move `json:"moves"`
@@ -23,7 +30,8 @@ type Query struct {
 	Rules Rules `json:"rules"`
 
 	// Komi is the komi to be used. Generally this is implied by the rules, so
-	// doesn't need to be set.
+	// doesn't need to be set. Note that the komi should be integers or half
+	// integers: ex: 8, 7.5, 6.5 etc.
 	Komi *float64 `json:"komi,omitempty"`
 
 	// BoardXSize is the width of the board
@@ -65,7 +73,7 @@ func NewQuery() *Query {
 
 // Move is a string-tuple having the form [<MOVE>, <POS>].
 // For example : ["B", "C4"]
-type Move []string
+type Move [2]string
 
 // Rules are rule-alias strings.
 type Rules string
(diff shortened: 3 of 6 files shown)


core/point/point.go
```diff
@@ -1,13 +1,40 @@
...
+	21: 'v', 22: 'w', 23: 'x', 24: 'y', 25: 'z', 26: 'A', 27: 'B',
+	28: 'C', 29: 'D', 30: 'E', 31: 'F', 32: 'G', 33: 'H', 34: 'I',
+	35: 'J', 36: 'K', 37: 'L', 38: 'M', 39: 'N', 40: 'O', 41: 'P',
+	42: 'Q', 43: 'R', 44: 'S', 45: 'T', 46: 'U', 47: 'V', 48: 'W',
+	49: 'X', 50: 'Y', 51: 'Z',
+}
+
+// sgfToPointRef is an anonymous function that runs and inverts the
+// pointToSgfMap to create the reverse translation reference
+var sgfToPointRef = func(mapIn map[int64]rune) map[rune]int64 {
```
artasparks: in general, in go, you shouldn't name things blahMap or blahSlice. That embeds the type in the variable name - but that's unnecessary.

so instead of blahMap just use blah, instead of mapIn, mapElem, you even can just use `in`, `out`.

core/point/point.go
```diff
@@ -1,13 +1,40 @@
...
 // Point is a basic point. Although simple, the member variables are kept
 // private to ensure that Point remains immutable.
 type Point struct {
 	x int64
 	y int64
 }
 
+// pointToSgfMap is a translation reference between int64 Point
+// and string SGF-Point (rune) values
+var pointToSgfMap = map[int64]rune{
```
artasparks: You shouldn't have this diff. I think you need to

```
git fetch upstream
git merge upstream/master
```

If you don't have upstream set up, you'll need to

```
git remote add upstream git@github.com:otrego/clamshell.git
```

core/point/point_test.go
```diff
@@ -13,3 +13,114 @@ func TestCreate(t *testing.T) {
...
+func TestPointToSGFTranslate(t *testing.T) {
+	// First test translation from integer-point to SGF-string-point
+	testToSGFCases := []struct {
+		desc string
+		in   *Point
+		want string
+	}{
+		{
+			desc: "Point => SGF",
+			in: &Point{
```
artasparks: just use New(...). Here and below

core/point/point_test.go
```diff
@@ -13,3 +13,114 @@ func TestCreate(t *testing.T) {
...
+func TestSGFToPointTranslate(t *testing.T) {
+	testToPointCases := []struct {
+		desc string
+		in   string
+		want *Point
+	}{
+		{
+			desc: "SGF => Point",
+			in:   "iq",
+			want: &Point{
```
artasparks: just use New. Here and below

core/point/point_test.go
```diff
@@ -13,3 +13,114 @@ func TestCreate(t *testing.T) {
 		t.Errorf("p.Y()=%v, expected %v", gotY, 2)
 	}
 }
+
+func TestPointToSGFTranslate(t *testing.T) {
+	// First test translation from integer-point to SGF-string-point
+	testToSGFCases := []struct {
+		desc string
+		in   *Point
+		want string
```
artasparks: should have expErrSubstr or similar. See: https://github.com/otrego/clamshell/blob/master/core/sgf/parser_test.go#L23

core/point/point_test.go
```diff
@@ -13,3 +13,114 @@ func TestCreate(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestSGFToPointTranslate(t *testing.T) {
+	testToPointCases := []struct {
+		desc string
+		in   string
+		want *Point
```
artasparks: should have expErrSubstr or similar

core/point/point.go
```diff
@@ -17,7 +44,48 @@ func New(x, y int64) *Point {
...
+// (struct with two rune/char values).
+func NewFromSGF(sgfPt string) (*Point, error) {
+	var intX int64
+	var intY int64
+	var err02 = fmt.Errorf("")
+	if (sgfPt != "") && (len(sgfPt) == 2) {
+		intX = sgfToPointRef[rune(sgfPt[0])]
+		intY = sgfToPointRef[rune(sgfPt[1])]
+		err02 = nil
+	} else {
```
artasparks: you don't need the else block.

core/point/point.go
```diff
@@ -17,7 +44,48 @@ func New(x, y int64) *Point {
...
+			"and less than or equal to 51. ")
+	}
+	return sgfOut, err01
+}
+
+// NewFromSGF converts an SGF point (
+// two letter string) to a pointer-type (immutable) *Point (
+// (struct with two rune/char values).
+func NewFromSGF(sgfPt string) (*Point, error) {
+	var intX int64
```
artasparks: You don't need to define these vars in the parent scope. See comments below.

core/point/point.go
```diff
@@ -17,7 +44,48 @@ func New(x, y int64) *Point {
...
+	var intY int64
+	var err02 = fmt.Errorf("")
+	if (sgfPt != "") && (len(sgfPt) == 2) {
```
artasparks: For this code block, you can do just

```
return New(intX, intY, nil)
```

However, the logic should be reversed.

```golang
if ...check error cases... {
  return nil, fmt.Errorf("some error")
}
do non-error stuff
```

core/point/point.go
```diff
@@ -17,7 +44,48 @@ func New(x, y int64) *Point {
...
-func (p *Point) Y() int64 { return p.y }
+func (pt *Point) Y() int64 { return pt.y }
+
+// ToSGF converts a pointer-type (immutable) *Point to an SGF Point (
+// two letter string).
+// The returned value is 0-indexed.
+func (pt *Point) ToSGF() (string, error) {
+	var sgfOut string
+	var err01 = fmt.Errorf("")
+	if (0 <= pt.X()) && (pt.X() <= 51) && (0 <= pt.Y()) && (pt.Y() <= 51) {
```
artasparks: See my comments below. General format for error cases should be

```golang
if ...check error cases... {
  return nil, fmt.Errorf("...")
}
return blah, nil

File: core/point/point.go [hunk 11]
```diff
@@ -17,7 +44,48 @@ func New(x, y int64) *Point {
...
 
 // Y returns the y-value.
-func (p *Point) Y() int64 { return p.y }
+func (pt *Point) Y() int64 { return pt.y }
+
+// ToSGF converts a pointer-type (immutable) *Point to an SGF Point (
+// two letter string).
+// The returned value is 0-indexed.
+func (pt *Point) ToSGF() (string, error) {
+	var sgfOut string
```
artasparks: not necessary to define parent scope. Just have returns in the nested blocks.

artasparks (commented):
Since this is a clean up, I'm going to be very nit-picky =)

core/point/point_test.go
```diff
@@ -13,3 +13,114 @@ func TestCreate(t *testing.T) {
...
+			want: &Point{
+				x: 40,
+				y: 51,
+			},
+		},
+	}
+
+	for _, tc := range testToPointCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			toPointOut, _ := NewFromSGF(tc.in)
```
artasparks: nit: just out instead of toPointOut

core/point/point_test.go
```diff
@@ -13,3 +13,114 @@ func TestCreate(t *testing.T) {
...
+			// Utilizing the point.go *Point type X Y getters below
+			pointX := toPointOut.X()
```
artasparks: you can just have x := toPointOut.X(). Or, honestly, you probably don't need this variable. blah.X() is pretty short.

PR #106, opened 2020-10-03
KiloLimaFoxtrot Cleaned Up Point Class

### LINKED ISSUE #93: Clean up Point Class
There's a couple of cleanups we talked about previously for the point class in https://github.com/otrego/clamshell/pull/79

- Inverse method for map
- Check error cases in unit tests (missing >= 0 in https://github.com/otrego/clamshell/blob/master/core/point/point.go#L61)
- There's one error with edge casses (negative points)
- Remove commented code

KiloLimaFoxtrot:
Cleaned up the Point Class by implementing an inverse method that serves as a reference to translate from SGF to Point. Minimized and condensed code lines, and removed relics from old code. The Point Test error substring is possibly incomplete, need clarification on that functionality. Code from our new Italian friend ilmanzo remains. Also created nested return blocks in translation methods to reverse and correct the error check logic path. 

Fixes: https://github.com/otrego/clamshell/issues/93


KiloLimaFoxtrot:
Added a point test error sub-string initial structure. Placed a slot for an error sub-string in the test structs, and one commented out error test case. Will need some additional clarification to complete the error sub-string.

core/point/point.go
```diff
@@ -52,49 +49,39 @@ func (pt *Point) X() int64 { return pt.x }
...
 	}
-	return sgfOut, err01
+	return string(pointToSgfRef[pt.X()]) + string(
+		pointToSgfRef[pt.Y()]), nil
+
 }
 
-// String() method to represent and print a Point, useful for debugging and test purposes
+// The below function and related elements were contributed by
+// un nuovo amico GitHub dall'Italia, di nome ilmanzo
```
artasparks: This comment isn't necessary. you can remove.

core/point/point_test.go
```diff
@@ -17,121 +17,107 @@ func TestCreate(t *testing.T) {
...
 		{
 			desc: "Point => SGF",
-			in: &Point{
-				x: 40,
-				y: 51,
-			},
+			in:   New(40, 51),
 			want: "OZ",
 		},
+		// // Error Cases
```
artasparks: remove? or check error cases?

artasparks: To check error cases, the general pattern I use is

```
val, err := someMethod(blah)
cerr := errcheck.CheckCases(err, tc.expErrSubstr)
if cerr != nil {
	t.Fatal(cerr)
}
if err != nil {
	return
}
```

artasparks (dismissed):
LGTM. Good cleanups. A good idea to check the error cases, if you have the time.

artasparks:
fixes: https://github.com/otrego/clamshell/issues/93

core/errcheck/errcheck.go
```diff
@@ -9,11 +9,15 @@ import (
 // CheckCases checks error cases for tests
 func CheckCases(err error, expErrSubstr string) error {
 	if err == nil && expErrSubstr != "" {
-		return fmt.Errorf("got no error but expected error containing %q", expErrSubstr)
+		return fmt.Errorf("got no error but expected error containing %q",
```
artasparks, line 12: Go doesn't wrapping strings, "go has no line length limit": https://golang.org/doc/effective_go.html#formatting

core/point/point.go
```diff
@@ -46,55 +43,46 @@ func New(x, y int64) *Point {
 	}
 }
 
-// X returns the x-value.
+// X method receives a pointer argument of type *Point,
```
artasparks: This is a little wordy. The types gives much of the information. Just something like the following would be sufficient:

`// X returns the x-value.`

core/point/point.go
```diff
@@ -46,55 +43,46 @@ func New(x, y int64) *Point {
...
-		sgfOut = sgfX + sgfY
-		err01 = nil
-	} else {
-		sgfOut = ""
-		err01 = fmt.Errorf("*Point int64 x and y value entries must" +
-			" be greater than or equal to 0, " +
-			"and less than or equal to 51. ")
-		// err01 = errors.New("*Point x and y value entries must be" +
-		// 	" greater than or equal to 0, and less than or equal to 51")
+	if (pt.X() < 0) || (51 < pt.X()) || (pt.Y() < 0) || 51 < (pt.Y()) {
```
artasparks, line 55: Won't this always error-out now?

core/point/point.go
```diff
@@ -46,55 +43,46 @@ func New(x, y int64) *Point {
 	}
 }
 
-// X returns the x-value.
+// X method receives a pointer argument of type *Point,
+// and returns the x-value.
 func (pt *Point) X() int64 { return pt.x }
 
-// Y returns the y-value.
+// Y method receives a pointer argument of type *Point,
```
artasparks: Same comment here about being a bit too wordy in the comments

core/point/point.go
```diff
@@ -46,55 +43,46 @@ func New(x, y int64) *Point {
...
+// and returns the y-value.
 func (pt *Point) Y() int64 { return pt.y }
 
-// ToSGF converts a pointer-type (immutable) *Point
+// ToSGF method receives a pointer argument of type *Point,
```
artasparks: Same comment here about being a bit too wordy in the comments.

core/point/point.go
```diff
@@ -46,55 +43,46 @@ func New(x, y int64) *Point {
...
+				"and less than or equal to 51. ")
 	}
-	return sgfOut, err01
+	return string(pointToSgfRef[pt.X()]) + string(
+		pointToSgfRef[pt.Y()]), nil
+
 }
 
-// String() method to represent and print a Point, useful for debugging and test purposes
+// String method receives a pointer argument of type Point,
```
artasparks: Again, a bit too wordy. You don't need to describe what's already clear from the type signature.

core/treepath/treepath.go
```diff
@@ -78,11 +78,13 @@ func Parse(path string) (Treepath, error) {
```
artasparks, line 78: Just revert this file. These look like line-length changes / spacing changes.

core/errcheck/errcheck.go
```diff
@@ -9,11 +9,15 @@ import (
 // CheckCases checks error cases for tests
```
artasparks, line 9: I would revert this file -- these are just line-length changes.

artasparks: Here's a SO article that might help: https://stackoverflow.com/questions/215718/how-can-i-reset-or-revert-a-file-to-a-specific-revision

core/point/point_test.go
```diff
@@ -17,121 +18,168 @@ func TestCreate(t *testing.T) {
 	}
 }
 
+// The below function TestPointToString() was authored by ilmanzo
```
artasparks, line 21: rm this. It doesn't matter who commits were authored by.

core/sgf/parser_test.go
```diff
@@ -250,10 +250,12 @@ AB[na][ra][mb][rb][lc][qc][ld][od][qd][le][pe][qe][mf][nf][of][pg]
 				for prop, expData := range pmap {
```
artasparks, line 250: Revert this file -- this looks like purely line-length changes.

core/point/point_test.go
```diff
@@ -17,121 +18,168 @@ func TestCreate(t *testing.T) {
...
+			expErrSubstr: "Point Y value > 51, out of range",
+		},
 	}
 
 	for _, tc := range testToSGFCases {
 		t.Run(tc.desc, func(t *testing.T) {
-			toSGFOut, _ := New(tc.in.x, tc.in.y).ToSGF()
-			if toSGFOut != tc.want {
+			sgfOut, err := New(tc.in.x, tc.in.y).ToSGF()
+			cerr := errcheck.CheckCases(err, tc.expErrSubstr)
```
artasparks, line 87: first check the error cases

core/point/point_test.go
```diff
@@ -17,121 +18,168 @@ func TestCreate(t *testing.T) {
...
-				t.Errorf("%q.toPointOut.x = %q, but wanted %q", tc.in,
-					pointX, tc.want.x)
-			}
-			if pointY != tc.want.y {
-				t.Errorf("%q.toPointOut.y = %q, but wanted %q", tc.in,
-					pointY, tc.want.y)
+			pntOut, err := NewFromSGF(tc.in)
+			cerr := errcheck.CheckCases(err, tc.expErrSubstr)
+			// Utilizing the point.go *Point type X Y getters below
+			if pntOut != tc.want {
```
artasparks, line 159: This is too indented (see above comment)

artasparks, line 159: You can't compare points this way since points are Pointer types. you have to compare the x & y values (or make an equals method). This will just compare the actual address, and will always be false.

core/point/point_test.go
```diff
@@ -17,121 +18,168 @@ func TestCreate(t *testing.T) {
...
+		},
 	}
 
 	for _, tc := range testToSGFCases {
 		t.Run(tc.desc, func(t *testing.T) {
-			toSGFOut, _ := New(tc.in.x, tc.in.y).ToSGF()
-			if toSGFOut != tc.want {
+			sgfOut, err := New(tc.in.x, tc.in.y).ToSGF()
+			cerr := errcheck.CheckCases(err, tc.expErrSubstr)
+			if sgfOut != tc.want {
```
artasparks, line 88: this is too indented. To avoid excessive indenting, break in case of error
```
if blah {
  t.Errorf(...)
  return // break
}
// continue on
```

artasparks, line 88: This won't accurately catch error cases.

In the error cases, it will always be the case that out ("") == expected ("") and it will evaluate to true.

core/point/point.go
```diff
@@ -46,55 +43,46 @@ func New(x, y int64) *Point {
...
+// and performs to represent and print a Point;
+// useful for debugging and test purposes - ilmanzo contributing
 func (pt Point) String() string {
 	return fmt.Sprintf("{%d,%d}", pt.x, pt.y)
 }
 
-// NewFromSGF converts an SGF point (two letter string)
-// to a pointer-type (immutable) *Point.
+// NewFromSGF function receives a pointer argument of type string,
```
artasparks, line 71: Again, a bit too wordy.

core/point/point.go
```diff
@@ -46,55 +43,46 @@ func New(x, y int64) *Point {
...
-		sgfOut = sgfX + sgfY
-		err01 = nil
-	} else {
-		sgfOut = ""
-		err01 = fmt.Errorf("*Point int64 x and y value entries must" +
-			" be greater than or equal to 0, " +
-			"and less than or equal to 51. ")
-		// err01 = errors.New("*Point x and y value entries must be" +
-		// 	" greater than or equal to 0, and less than or equal to 51")
+	if (pt.X() < 0) || (51 < pt.X()) || (pt.Y() < 0) || 51 < (pt.Y()) {
```
artasparks, line 55: You have too many parentheses. `<` has a higher precedence than `||` so you don't need the sub grouping: https://golang.org/ref/spec#Operator_precedence

core/point/point.go
```diff
@@ -46,55 +43,46 @@ func New(x, y int64) *Point {
...
-	if (sgfPt != "") && (sgfPt != "--") && (len(sgfPt) == 2) {
-		intX = sgfToPointMap[rune(sgfPt[0])]
-		intY = sgfToPointMap[rune(sgfPt[1])]
-		err02 = nil
-	} else {
-		intX = 99
-		intY = 99
-		err02 = fmt.Errorf("SGF string x and y value entries must non" +
-			"-empty and of length 2 (runes/chars). ")
+	if (sgfPt == "") || (len(sgfPt) != 2) {
```
artasparks, line 74: Similar comment to above. `==` has higher precedence than `||`, so you don't need the parentheses

core/point/point.go
KiloLimaFoxtrot: Hey, I don't think so because with the previous conjunctive case, ie with &&, it will only error out with both bas conditions met; such as (pt.X() < 0) && (51 < pt.X()), but with the above disjunctive, ie with ||, it will error out with either bad condition.(pt.X() < 0) || (51 < pt.X()), and same with Y().. Does that make sense to you? Let me know if I am missing something.  :  )

core/point/point_test.go
KiloLimaFoxtrot: I actually do know that, and just missed that honestly. Also, I am having some trouble making this test cases method work, maybe I need to revisit the mechanics? And I've been very busy with school. Thanks for being flexible.

core/point/point.go
artasparks: oh, I see. I would recommend keeping consistent pattern for the conditionals -- I.e., Blah <compare> Number. `if pt.X() < 0 || pt.X() >51 || pt.Y() < 0 || pt.Y() > 56`

artasparks:
This is pretty old, so I'm going to close. Feel free to re-initiate pull request with updated code =)

PR #111, opened 2020-10-07
added function to convert the treepath to a string

TrevorPeters:


core/treepath/treepath.go
```diff
@@ -177,3 +177,16 @@ func (tp Treepath) Apply(n *game.Node) *game.Node {
 	}
 	return curNode
 }
+
+// Returns the treepath as a string.
```
artasparks: Format for comments is generallly `// <MethodName> does blah`

ex:

```go
// String returns the treepath as a string
```

core/treepath/treepath.go
```diff
@@ -177,3 +177,16 @@ func (tp Treepath) Apply(n *game.Node) *game.Node {
 	}
 	return curNode
 }
+
+// Returns the treepath as a string. 
+// examples: 
+// 	[]				becomes "[]"
+// 	[1]				becomes	"[1]"
+// 	[1,2,0,2,2,2] 	becomes "[1,2,0,2,2,2]"
+func (tp Treepath) String() string {
```
artasparks, line 186: Note: I think we want to string methods.

`String()`, which is this method and `CompactString()` which returns the shortened version with the `:` character. Ex: `CompactString([0,0,0,0]) => "0:4"`

I would do that as a follow-up PR.

core/treepath/treepath_test.go
```diff
@@ -171,3 +171,21 @@ func TestApplyPath(t *testing.T) {
 		})
 	}
 }
+
+func TestString(t *testing.T){
```
artasparks: Typically way to format this would be with a table test: https://github.com/golang/go/wiki/TableDrivenTests

See above for some examples. Also, here's one:

```go
testCases := []struct{
  desc string
  in []int
  out string
}{
...
}
```

core/treepath/treepath.go
```diff
@@ -177,3 +177,16 @@ func (tp Treepath) Apply(n *game.Node) *game.Node {
...
+	var strArr []string
+	for _, i := range tp{
```
artasparks: formatting is a bit off. gofmt will automatically fix this, but it needs to be configured. What editor are you using?

core/treepath/treepath.go
```diff
@@ -177,3 +177,16 @@ func (tp Treepath) Apply(n *game.Node) *game.Node {
...
+		strArr = append(strArr, strconv.Itoa(i))
+	}
+	return "[" + strings.Join(strArr, ",") + "]"
```
artasparks: Prefer fmt.Sprintf(...) instead of manually constructing strings. I think fmt.Sprintf("%v", tp) might be actually what you want here, though =)

artasparks (commented):
Awesome! Thanks for the PR!

core/treepath/treepath.go
TrevorPeters: VScode

artasparks: I would check out: https://code.visualstudio.com/docs/languages/go. Especially enable the autoformatting / linting / vet to start with =)

core/treepath/treepath.go
TrevorPeters: fmt.Sprintf("%v", tp) will cause StackOverflow. I can change the return to fmt.Sprintf("%v", strArr) though.

artasparks: oh, lol. yes. Because %v will implicitly use String()

PR #114, opened 2020-10-10
96 command wrapper

### LINKED ISSUE #96: Add a command-wrapper for katago
To run analyses from `kanalyze`, we need to start `katago` from `kanalyze` and then connect stdin/stdout/stderr pipes. 

https://golang.org/pkg/os/exec/ should do what we want, but I haven't been able to make it work. I have the beginnings of such a thing in https://github.com/otrego/clamshell/pull/95

mhoak:
In short, this commit adds an initial wrapper around Katago that handles the stdin, stdout, stderr well.

It will need to be refactored a bit so it can be tested, but I wanted to get this in as-is for review. 

Fixes: https://github.com/otrego/clamshell/issues/96


.githooks/pre-commit
```diff
@@ -49,7 +49,7 @@ do
   # Run govet on the staged file and check the exit status
   go vet $FILE
   if [[ $? != 0 ]]; then
-    printf "\t$go vet $FILE ${TERM_BLACK}${TERM_BACKGROUND_RED}FAILED{TERM_END_COLOR}\n"
```
artasparks, line 52: ah, this isn't quite right `go vet` should work on the package, rather than the file.

artasparks: We should also update the docs --  I think @TrevorPeters ran into this.

(also, aren't githooks per-user? I'm surprised this is in the PR)

go.mod
```diff
@@ -3,13 +3,65 @@ module github.com/otrego/clamshell
 go 1.14
 
 require (
```
artasparks, line 5: where are all these deps coming from??

core/katago/katago.go
```diff
@@ -19,33 +26,148 @@ type Analyzer struct {
...
+
+		analyzerSingleton.bootWait.Add(1)
+		err = analyzerSingleton.start()
+	})
+	return analyzerSingleton, err
+}
+
+// AnalyzeGame is a synchronous request for Katago to process a game
+func (an *Analyzer) AnalyzeGame(json string) (*string, error) {
+	an.bootWait.Wait()
```
artasparks, line 85: I don't think you need to wait for katago to start in order to send requests. So, seems like we can elide this?

core/katago/katago.go
```diff
@@ -19,33 +26,148 @@ type Analyzer struct {
...
+	an.stdinWrite <- json + "\n"
```
artasparks: probably only add "\n" if it's not already there

core/katago/katago.go
```diff
@@ -19,33 +26,148 @@ type Analyzer struct {
...
-// Cmd creates the Katago analysis command.
-func (an *Analyzer) Cmd() *exec.Cmd {
-	threads := an.AnalysisThreads
-	if threads == 0 {
-		threads = 8
+var analyzerSingleton *Analyzer
+var once sync.Once
+
+// GetAnalyzer returns the Katago Analyzer Singleton
+func GetAnalyzer(model string, configPath string, numThreads int) (*Analyzer, error) {
```
artasparks: I think it would be better to separate the steps a bit:

1. Create the Analyzer
2. Start the Analyzer
3. Stop the Analyzer.

Otherwise, you'll only ever be able to: Start katago & Analyze exactly one file

core/katago/katago.go
```diff
@@ -19,33 +26,148 @@ type Analyzer struct {
 
 	// Number of analysis threads. Defaults to 16 if not specified.
 	AnalysisThreads int
+
+	cmd          *exec.Cmd
```
artasparks: Right now, this pattern creates a lot of shared state and means you can't concurrently call the analyzer. I.e., you cannot currently do

```
an, _ := GetAnalyzer(...)
res1 := an.AnalyzeGame(foo)
res2 := an.AnalyzeGame(bar)
```

This would be a big overhaul, so I can see this being a separate PR.

core/katago/katago.go
```diff
@@ -19,33 +26,148 @@ type Analyzer struct {
...
+	return nil
+}
+
+// Cmd creates the Katago analysis command.
+func (an *Analyzer) createCmd() {
+	threads := an.AnalysisThreads
+	an.cmd = exec.Command("katago", "analysis", "-model", an.Model, "-config", an.Config, "-analysis-threads", strconv.Itoa(threads))
+}
+
+func (an *Analyzer) startAnalyzerIO() error {
```
artasparks, line 214: I would make this public and leave it up to the user to call. That's not compatible with the current design, so probably a separate PR.

core/katago/katago.go
```diff
@@ -19,33 +26,148 @@ type Analyzer struct {
...
+	})
+	return analyzerSingleton, err
+}
+
+// AnalyzeGame is a synchronous request for Katago to process a game
+func (an *Analyzer) AnalyzeGame(json string) (*string, error) {
+	an.bootWait.Wait()
+	an.stdinWrite <- json + "\n"
+	output := <-an.katagoOutput
+	if strings.Contains(output, "error") {
```
artasparks: Hmm. Why would the STDOUT stream have the words "error"?

core/katago/katago.go
```diff
@@ -19,33 +26,148 @@ type Analyzer struct {
...
+		glog.Warningf("Katago had an error: %v", output)
+		return nil, errors.New(output)
 	}
-	return exec.Command("katago", "analysis", "-model", an.Model, "-config", an.Config, "-analysis-threads", strconv.Itoa(threads))
+	return &output, nil
```
artasparks: why pass back a pointer?

go.mod
artasparks: Can you run `go mod tidy`?

artasparks (commented):
Meta comment: I think it's fine to get this in and iterate, but there are definitely things we should improve. 

1. Do you want to polish this or submit it? 
2. Do you want to own follow-up work or pass it off?

artasparks:
Branch has conflicts. Can you merge in upstream/master?

artasparks (approved):
I'd like to get this in so I can iterate, even though there are some failures on head. I'll send some fixups =)

PR #120, opened 2020-10-19
Several improvements to katago/katalyze

### LINKED ISSUE #119: Change logging verbosity in katago/katalyze
Right now the logging in `katago.go` is very verbose. We should use leveled logging to make it configurable, but quieter by default.

artasparks:
- Reduce log level to improve readibility
- Use a more understandable game
- Refactor katago to not be a singleton
- Refactor katago have less indentation.

Fixes: #119


core/katago/katago.go
```diff
@@ -48,49 +46,59 @@ type Analyzer struct {
 	bootWait sync.WaitGroup
 }
 
-var analyzerSingleton *Analyzer
-var once sync.Once
+// New creates a new analyzer. Before being used, the Start() must be called.
+func New(model string, configPath string, analysisThreads int) *Analyzer {
```
mhoak, line 50: This is much better in the non-singleton world; we dont need singletons so this is the right choice.

core/katago/katago.go
```diff
@@ -130,102 +140,84 @@ func (an *Analyzer) readCollectedResult(analysisID string) *AnalysisList {
...
 // Cmd creates the Katago analysis command.
-func (an *Analyzer) createCmd() {
-	threads := an.AnalysisThreads
-	an.cmd = exec.Command("katago", "analysis", "-model", an.Model, "-config", an.Config, "-analysis-threads", strconv.Itoa(threads))
+func (an *Analyzer) createCmd() *exec.Cmd {
+	return exec.Command("katago", "analysis", "-model", an.Model, "-config", an.Config, "-analysis-threads", strconv.Itoa(an.AnalysisThreads))
 }
 
-func (an *Analyzer) startAnalyzerIO() error {
+func (an *Analyzer) startAnalyzerIO(c *exec.Cmd) error {
```
mhoak: Why pass in the command vs referencing the command in the Analyzer? 
It _seems_ like this is a good practice because then you don't have to reason about the state of the Analyzer's command - there just is a command. Is that the right way to think about this?

mhoak (approved):
Overall looks good. There are a few things I don't understand. Pinging you separately about those.

core/katago/katago.go
artasparks: Ack

core/katago/katago.go
artasparks: I was trying to reduce the amount of state that was being stored in the Analyzer --> Anything that's in the analyzer struct seems like it should live for the life of the object. The seems like it isn't like that -- it is used to things off and then we don't use the cmd again, since we quit by closing the STDIN Pipe.

I'm not tied to this change -- really just a matter of taste.

artasparks: Reverted this change.

mhoak (approved):

PR #134, opened 2020-11-12
Convert Game to SGF (Issue#127)

TrevorPeters:


TrevorPeters:
Interesting. the test passes when I run it in Vs code, but fails when I run it in the command line. will investigate further.

Update: the test passes and fails randomly. Vs code cached the first successful result leading me to believe that it was consistently passing. I need to change the testing method

Update 2: testing issue resolved :D

core/sgf/parser.go
```diff
@@ -395,3 +395,38 @@ func postProcessProperties(n *game.Node, prop string, propData []string) error {
 
 	return nil
 }
+
+// Serialize converts a Game into SGF format.
+// Calls serializeHelper.
```
artasparks, line 400: nit: You don't need to mentioned the internals (serializeHelper) in the public docs.

core/sgf/parser.go
```diff
@@ -395,3 +395,38 @@ func postProcessProperties(n *game.Node, prop string, propData []string) error {
...
+func Serialize(g *game.Game) string {
+	return fmt.Sprintf("(%s)", serializeHelper(g.Root))
+}
+
+// serializHelper is a recursive DFS searching all
+// descendant nodes of n.
+func serializeHelper(n *game.Node) string {
```
artasparks, line 407: This is very slick, nice.

artasparks (approved):
Nice!

PR #139, opened 2020-11-21
Added Codeowners to Clamshell.

nelsonjhk:
Github allows for auto-assigning code owners to relevant codebases. This file assigns maintainers to pull requests for final review.


.github/CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+# Codeowners file for auto-assignment of experienced developers for code review
+
+*	@kashomon @mhoak
+
+# Auto-assignment of relevant developers for code review
+
+core/sgf/	@kashomon @mhoak @TrevorPeters
```
artasparks, line 7: Are directory assignment rules inherited? 

I.e., does this also get the * rules by default? In which case we could just put TrevorPeters?

artasparks (approved):

artasparks:
Fixes #140

.github/CODEOWNERS
nelsonjhk: My understanding is that the lowest level takes precedence, in which case only TrevorPeters would be assigned.

artasparks:
LGTM.

PR #142, opened 2020-11-21
Added basic features for points

jonathanmli:


core/point/point.go
```diff
@@ -12,6 +12,12 @@ type Point struct {
 	y int64
 }
 
+// Coord is a float point for representing points in space
+type Coord struct {
```
artasparks: How about we move this to core/point/coord.go? Would make it a bit clearer, I think.

core/point/point.go
```diff
@@ -52,6 +58,20 @@ func (pt *Point) X() int64 { return pt.x }
 // Y returns the y-value.
 func (pt *Point) Y() int64 { return pt.y }
 
+// NewCd creates a new immutable Coord.
+func NewCd(x, y float64) *Coord {
```
artasparks: Just NewCoord I think

artasparks:
Let me know when you move the Coord and we can merge the PR =)

artasparks:
ping =)

jonathanmli:
done! (I think)

artasparks:
I'm not seeing the new file. Check your `git status` -- I suspect there's some untracked files. So, I think you might need to do a `git add -A` and then `git commit` or `git commit -m <msg>`

artasparks (approved):
Looks good to me!

PR #144, opened 2020-11-22
Move treepath into Game

### LINKED ISSUE #135: Move treepath to game
We probably should move `core/treepath` to `core/game`. There's a lot of overlap in functionality between the Node and Treepath. For example, it would be useful to have a game take a treepath to

1. make a flattened copy to that point
2. move to that location.

etc.

artasparks:
Treepath is intimately tied with 'game` with is really more of a movetree. So, it makes the most sense to me to move it into the game package.

As a follow up, I think that `game` should be renamed `movetree` or similar. 

Fixes #135

### DIFF
--- a/core/game/treepath_test.go
+++ b/core/game/treepath_test.go
@@ -1,10 +1,11 @@
-package game
+package game_test
 
 import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/otrego/clamshell/core/errcheck"
+	"github.com/otrego/clamshell/core/game"
 	"github.com/otrego/clamshell/core/sgf"
 )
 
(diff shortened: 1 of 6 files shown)


TrevorPeters (approved):

core/game/treepath_test.go
```diff
@@ -1,14 +1,15 @@
-package treepath
+package game_test
```
mhoak, line 1: Shouldn't this be package game?

mhoak (approved):

core/game/treepath_test.go
artasparks: It's intentionally not. This depends on SGF (for convenience), and SGF depends on game. Using game_test (separate package for tests), fixes the cycle.

PR #157, opened 2020-11-28
Adds initial storage code.

mhoak:
Adds an initial storage file-store. I think this is a decent approach as it treats the storage as a pseudo database and should be able to grow as the needs evolve.


core/storage/diskstore.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Put is method to Put a file to disk
+func (ds *DiskStore) Put(t StoredDataType, filename string, json string) error {
+	path := ds.path(t, filename)
+	return ioutil.WriteFile(path, []byte(json), os.ModePerm)
+}
+
+func (ds *DiskStore) path(t StoredDataType, filename string) string {
+	return fmt.Sprintf("%s/%s/%s", ds.rootDir, t, filename)
```
artasparks: Prefer path.Join for system compatibility

core/storage/diskstore.go
```diff
@@ -0,0 +1,78 @@
...
+func (ds *DiskStore) ensureDirectoryStructure(rootDir string) error {
+	fileInfo, err := os.Stat(rootDir)
+	if err != nil {
+		return err
+	}
+	if !fileInfo.IsDir() {
+		return errors.New("Directory does not exist")
+	}
+	for _, t := range storedDataTypes {
+		curDir := fmt.Sprintf("%s/%s", ds.rootDir, t)
```
artasparks: here also: path.Join

core/storage/diskstore.go
```diff
@@ -0,0 +1,78 @@
...
+		fileInfo, err := os.Stat(curDir)
+		if err != nil {
+			os.Mkdir(curDir, 0755)
```
artasparks: nit: Maybe somewhere having a 

```go
const defaultDirPrms = 0755
```

core/storage/diskstore.go
```diff
@@ -0,0 +1,78 @@
...
+
+// ensureDirectoryStructure ensures that paths exist for each
+// of the expected file outputs
+func (ds *DiskStore) ensureDirectoryStructure(rootDir string) error {
+	fileInfo, err := os.Stat(rootDir)
+	if err != nil {
+		return err
+	}
+	if !fileInfo.IsDir() {
+		return errors.New("Directory does not exist")
```
artasparks: nit: errors usually are not capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

core/storage/diskstore_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"io/ioutil"
+	"log"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestDiskGetNoFileErrors(t *testing.T) {
```
artasparks, line 16: nit: I get wary when I see real I/O in unit tests, even with tempdirs. I think it's better to mock out the I/O. That being said, this isn't a huge issue. Mocking out is a bit of work.

core/storage/filestore.go
```diff
@@ -0,0 +1,30 @@
+package storage
+
+// Filestore is a interface for file storage
+type Filestore interface {
```
artasparks, line 6: Let's put `ctx context.Context` as a parameter to all the methods. We'll need that for all network Filestores.

core/storage/filestore.go
```diff
@@ -0,0 +1,30 @@
...
+	Get(StoredDataType, string) (string, error)
+	List(StoredDataType, string) ([]string, error)
+	Put(StoredDataType, string, string) error
+}
+
+// StoredDataType designates which type of data is stored
+type StoredDataType string
+
+const (
+	// SGFInput storage directory
```
artasparks: These names seem a little too wordy to me. I've provided some suggestions to abbreviate.

core/storage/filestore.go
```diff
@@ -0,0 +1,30 @@
...
+	SGFInput StoredDataType = "input_sgf"
```
artasparks: how about just `games`

core/storage/filestore.go
```diff
@@ -0,0 +1,30 @@
...
+	// AnalysisOutput is the data type for katago output
+	AnalysisOutput StoredDataType = "katago_analysis"
```
artasparks: how about `analysis`?

core/storage/filestore.go
```diff
@@ -0,0 +1,30 @@
...
+	// AiProblemIdentification is the directory for AI identifed problems
+	AiProblemIdentification StoredDataType = "ai_problem_identification"
```
artasparks: I'm still not sure what this based on the name. Maybe `metadata`?

core/storage/filestore.go
```diff
@@ -0,0 +1,30 @@
...
+	// AiProblemSGF is the directory for AI problems that have been generated
+	// and turned into SGF.
+	AiProblemSGF StoredDataType = "ai_problem_sgf"
```
artasparks: How about just `problems`

core/storage/filestore.go
```diff
@@ -0,0 +1,30 @@
+package storage
```
artasparks, line 1: +package comment?

core/storage/diskstore_test.go
mhoak: Hrm. I'm not sure if that's worth doing. If you want it done, I'm happy to do it.

core/storage/filestore.go
mhoak: I'm not familiar with this specific pattern in go. I've added it as I think you meant it. Let me know.

core/storage/diskstore.go
```diff
@@ -0,0 +1,81 @@
...
+		rootDir: root,
+	}
+	err := ds.ensureDirectoryStructure(root)
+	if err != nil {
+		return nil, err
+	}
+	return ds, nil
+}
+
+// Get abstract method to get JSON content from a Filestore
```
artasparks: nit: content doesn't need to be JSON

core/storage/diskstore.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return "", err
+	}
+	return string(bytes), nil
+}
+
+// List is method to list available files
+func (ds *DiskStore) List(ctx context.Context, t StoredDataType) ([]string, error) {
+	ioutil.ReadDir(ds.rootDir)
+	return []string{""}, nil
```
artasparks: Looks like an error here -- doesn't seem to be doing a thing.

core/storage/filestore.go
artasparks: This is right.

core/storage/filestore.go
```diff
@@ -0,0 +1,29 @@
...
+type StoredDataType string
+
+const (
+	// Games storage directory
+	Games StoredDataType = "games"
+	// Analysis is the data type for katago output
+	Analysis StoredDataType = "analysis"
+	// Problems is the directory for AI problems that have been generated
+	// and turned into SGF.
+	Problems StoredDataType = "ai_problem_sgf"
```
artasparks: maybe just gen_problems? Or even just problems/?

core/storage/diskstore_test.go
artasparks: not a big deal

artasparks (approved):
LGTM

Issue #152, opened 2020-11-27
Add a new package for problems-logic


artasparks:
We should make a new package for problem-related logic. Probably `core/problems`

Issue #136, opened 2020-11-14
Take a treepath + game and produce a new flattened game


artasparks:
We want to: take a treepath + a game => a new game in a problem format.

`func Flatten([]Treepath, *Game) => []*Game`

artasparks:
So, a couple more details since this wasn't super well defined:. First, we should make a new problem-package: -- see #152 

After that, here's what I'm thinking:

- Roughly, the method should take a treepath, and apply the moves to a Board from `core/board`.
- Then, you'd extract the current board state and convert to the AW/BW moves (placements).
- The logic in #148 creates treepaths *after* the blunder. So, the way I see it, applying the treepath should produce a board at N-2 from the treepath. Then, you'd let the N-1 move as an actual move.

artasparks:
This is now done in #160. Thanks @TrevorPeters!

PR #160, opened 2020-11-29
added func to flatten a MoveTree

### LINKED ISSUE #158: Add a get-full-boardstate to Board
We should add a method like

```go
func (b *Board) GetFullBoardState() []*move.Move {
...
}
```

This would return the entire board state as a list of moves.

TrevorPeters:
fixes #158, #152, #136 

other changes
- fixed a few errors I encountered from core/board/board.go
- added a method to core/board/board.go that adds a stone without removing captured stones


core/board/board.go
```diff
@@ -31,18 +31,33 @@ func NewBoard(size int) *Board {
 }
 
 // AddStone adds a stone to the board.
+// Does not remove captured stones.
```
artasparks: I don't think we should have AddStone and PlaceStone. Just PlaceStone is sufficient -- by default captures should always be removed.

core/board/board.go
```diff
@@ -157,6 +174,23 @@ func (b *Board) getNeighbors(pt *point.Point) []*point.Point {
 	return points
 }
 
+// GetFullBoardState returns an array of all the current stone positions
+// as a []*move.Move .
```
artasparks: Nit: you don't need to specify the type in the comments because it's in the method signature.

core/problems/problems.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// tp ends one move after the blunder, so we follow the treepath
+	// to the move right before the blunder using len(tp) - 2;
+	for i := 0; i < len(tp)-2; i++ {
+		n = n.Next(tp[i])
+		if n == nil {
+			return nil, fmt.Errorf("treepath leads to nil movetree node")
+		}
+
+		_, err2 := b.PlaceStone(n.Move)
```
artasparks, line 53: I think it's worth a comment that you're intentionally discard captures here.

core/problems/problems.go
```diff
@@ -0,0 +1,94 @@
...
+	b, err := PopulateBoard(tp, g)
+	placements := b.GetFullBoardState()
+
+	if err != nil {
+		return nil, err
+	}
+
+	var sbSGF strings.Builder
+	var sbAB strings.Builder
+	var sbAW strings.Builder
```
artasparks: Ah, you're constructing an SGF manually. I would make edits on the actual node rather than string-construction it should be more robust.

TrevorPeters: Gotcha, should I use a new node, or can I manipulate the original movetree?

also, should the flat movetree point to the children of the node that we flatten to? I'm not sure if that's the way to communicate this.

do we want 
`flatTree.Root.Children = copy(originalTree.NodeFromTreepath(tp).Children)`
or do we want flatTree to have no children after Root

core/problems/problems.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+
+	gflat := movetree.New()
+	gflat.Root.Placements = b.GetFullBoardState()
+
+	aw := make([]string, 0)
+	ab := make([]string, 0)
+
+	for _, move := range gflat.Root.Placements {
```
artasparks: My work in https://github.com/otrego/clamshell/pull/161 should make this significantly simpler

core/problems/problems_test.go
```diff
@@ -0,0 +1,65 @@
...
+		desc string
+		tp   string
+		sgf  string
+	}{
+		{
+			desc: "entire game flatten",
+			tp:   "0:140",
+			sgf: `(;FF[4]CA[UTF-8]GM[1]DT[2020-08-05]PB[player1]PW[player2]
+				BR[2k] WR[5d] TM[259200]OT[86400 fischer] RE[W+R] SZ[19]
+				KM[6.5] RU[Japanese] ;B[pd] (;W[dc] (;B[qp] (;W[cq] (;B[np]
```
artasparks, line 24: I wouldn't try to simplify this game for a unit-test. As is, it
1. Has too many moves.
2. It has mega-unnecessary branching for some reason.

core/problems/problems_test.go
```diff
@@ -0,0 +1,75 @@
...
+				(;W[le] (;B[kd] (;W[ld] (;B[ji] (;W[ki] (;B[kj] (;W[jj]
+				(;B[ij] (;W[jk] (;B[nj] (;W[ee] (;B[jd] (;W[id] (;B[ie]
+				(;W[fe] )))))))))))))))))))))))))))))))))))))))))))))))
+				)))))))))))))))))))))))))))))))))))))))))))))))))))))))
+				))))))))))))))))))))))))))))))))))))`,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			tp, _ := movetree.ParsePath(tc.tp)
```
artasparks, line 51: In general, in a test, if there's an error, you should handle with t.Fatal(err) or t.Error(err)

core/problems/problems_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/otrego/clamshell/core/prop"
+	"github.com/otrego/clamshell/core/sgf"
+)
+
+func TestFlatten(t *testing.T) {
+
+	testCases := []struct {
+		desc string
+		tp   string
+		sgf  string
```
artasparks, line 17: I would put the expectations in the testcase -- i.e., the expected moves.

core/problems/problems.go
```diff
@@ -0,0 +1,59 @@
...
+func Flatten(tp movetree.Treepath, g *movetree.MoveTree) (*movetree.MoveTree, error) {
+	b, err := PopulateBoard(tp, g)
+	if err != nil {
+		return nil, err
+	}
+
+	gflat := movetree.New()
+	gflat.Root.Placements = b.GetFullBoardState()
+
+	for key, value := range g.Root.Properties {
```
artasparks, line 23: Ah, nice call here.

artasparks (commented):
Just some small fixes to the testcases, then looks good to merg.

artasparks:
Let's get this merged. I'd like to start using it =)

artasparks (approved):
I'll send a cleanup PR for some of the things I was mentioning.

PR #162, opened 2020-12-06
Rework docs to send most folks to Community page.

artasparks:
Moving more docs to community page. Namely, all of the CONTRIBUTING page and the katago-install stuff.


nelsonjhk (approved):
lgtm

PR #163, opened 2020-12-10
Connect everything together

artasparks:
This PR connects all the katalyze things together. Nothing is working perfectly =), but it is mostly working.

In particular, I:

- did some minor refactoring / fix-up of storage code.
- moved problem generation to a separate file
- created a shell script to automate running a smoketest

### DIFF
--- a/katalyze/main.go
+++ b/katalyze/main.go
@@ -90,10 +90,10 @@ func main() {
 	outDir := cwd
 	if *outputDir != "" {
 		outDir = path.Join(cwd, *outputDir)
-		if _, err := os.Stat(outDir); os.IsNotExist(err) {
+		if _, errz := os.Stat(outDir); os.IsNotExist(errz) {
 			os.Mkdir(outDir, 0775)
-		} else if err != nil {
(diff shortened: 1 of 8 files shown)


katalyze/main.go
```diff
@@ -70,80 +70,77 @@ func main() {
...
 	}
 
-	if err := process(files, an); err != nil {
-		glog.Exitf("error processing files: %v", err)
+	cwd, err := os.Getwd()
+	if err != nil {
+		glog.Exit(err)
 	}
+	outDir := cwd
+	if *outputDir != "" {
```
mhoak, line 91: Seems like it would make sense to use the DiskStorage here - but maybe you are trying to keep it decoupled from that.

mhoak (approved):

katalyze/main.go
artasparks: Right now disk-storage doesn't create the root directory, which is probably correct. So, I create it here if it doesn't exist. It does duplicate a bit what DiskStorage is doing, but it's probably the right separation of responsibilities.

PR #167, opened 2020-12-11
Some minor refactoring as as we move towards supporting cloudstorage …

mhoak:
…directly.

### DIFF
--- a/katalyze/main.go
+++ b/katalyze/main.go
@@ -40,6 +40,7 @@ import (
 )
 
 var (
+	storageProvider = flag.String("storage_provider", "", "Storage provider where analyzed games will be stored. Optional {localdisk|gcp}. Default is localdisk")
 	outputDir       = flag.String("output_dir", "", "Directory for returning the processed SGFs. By default, uses current directory")
 	bucket          = flag.String("bucket", "", "The bucket for CloudStorage of analyzed games. This param is optional.")
 	bucketPrefix    = flag.String("bucket_prefix", "", "Dictates the path within the bucket where games wil be stored. Required if the bucket parameter is supplied.")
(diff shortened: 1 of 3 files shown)


katalyze/main.go
```diff
@@ -41,6 +41,8 @@ import (
 
 var (
 	outputDir       = flag.String("output_dir", "", "Directory for returning the processed SGFs. By default, uses current directory")
+	bucket          = flag.String("bucket", "", "The bucket for CloudStorage of analyzed games. This param is optional.")
+	bucketPrefix    = flag.String("bucket_prefix", "", "Dictates the path within the bucket where games wil be stored. Required if the bucket parameter is supplied.")
```
artasparks, line 46: Not sure I understand bucket prefix?

katalyze/main.go
```diff
@@ -41,6 +41,8 @@ import (
 
 var (
 	outputDir       = flag.String("output_dir", "", "Directory for returning the processed SGFs. By default, uses current directory")
+	bucket          = flag.String("bucket", "", "The bucket for CloudStorage of analyzed games. This param is optional.")
```
artasparks, line 45: seems like you need a flag that's like `--storage-provider={localdisk|gcp}`

katalyze/main.go
mhoak: At least for AWS, you usually specify the bucket and the path within the bucket separately. Might be different for GCP

katalyze/main.go
mhoak: Good point. I was considering that. That does seem nicer.

katalyze/main.go
artasparks: Yeah, that's true for GCS also. maybe root folder?

artasparks: Or maybe prefix is canonical

mhoak: I think prefix is canonical

artasparks (approved):

TamaBrian:
Merry Christmas Message

From

    Brian and Etsuko Willis

PR #190, opened 2020-12-22
This is a short first-line.

### LINKED ISSUE #188: Add support for 9x9s in Katalyze
We should add support for 9x9s in Katalyze. It's not clear what will need to be fixed in order to process 9x9s, but we should add a 9x9 example to the testdata directory, so it can be run as part of the smoketest.

mhoak:
This is a really great test. Thanks for being here Lee, Rod, Josh, and Brian...
and anyone else.
It's important because GitHub uses the first line as the "pull request" subject by default.

Fixes: #188

### DIFF
--- a/katalyze/README.md
+++ b/katalyze/README.md
@@ -10,3 +10,5 @@ export KATAGO_MODEL=testdata/g170e-b10c128-s1141046784-d204142634.bin.gz
 export KATAGO_ANALYSIS_CONFIG=testdata/analysis_example.cfg
 go run main.go testdata/2019-04-30-1.sgf
 ```
+
+This is a test!


katalyze/README.md
```diff
@@ -10,3 +10,5 @@ export KATAGO_MODEL=testdata/g170e-b10c128-s1141046784-d204142634.bin.gz
 export KATAGO_ANALYSIS_CONFIG=testdata/analysis_example.cfg
 go run main.go testdata/2019-04-30-1.sgf
 ```
+
+This is a test!
```
mhoak, line 14: This seems like it was not intended for commit.

mhoak (commented):
This commit seems empty. Was it a mistake?

artasparks:
This is a test commit. =). Closing

PR #192, opened 2020-12-26
Komi Converter

### LINKED ISSUE #183: Add a converter for Komi
We should add a converter for the Komi property + fields in the `GameInfo` struct in node.go: https://github.com/otrego/clamshell/blob/master/core/movetree/node.go

Extra care will need to be paid to ensure that the decimal value of komi is only `.0` or `.5` -- ex: `6.5`. Also, Katago query conversion will need to change: https://github.com/otrego/clamshell/blob/master/core/katago/query.go#L180

* Property: https://www.red-bean.com/sgf/go.html#KM
* Property example: `KM[6.5]`

TrevorPeters:
Added Komi Converter to core/prop/converter_list.go

Fixes #183


artasparks (approved):
Looks good! good tests, good logic. Nice work.

PR #198, opened 2021-01-04
Add bbox and refactor crop package

### LINKED ISSUE #197: Add a bounding box struct
We should add a bounding box struct, which can be then used by the cropbox.

artasparks:
This adds a new bounding box struct and refactors the crop package to
use the bounding box struct.

Fixes: #197

### DIFF
--- a/core/bbox/bbox.go
+++ b/core/bbox/bbox.go
@@ -1,0 +1,53 @@
+// Package bbox provides utilities for managing various types of bounding boxes.
+package bbox
+
+import (
+	"fmt"
+
+	"github.com/otrego/clamshell/core/point"
+)
+
+// New creates a new bounding box.
+func New(tl, br *point.Point) (*BoundingBox, error) {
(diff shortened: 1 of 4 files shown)


core/bbox/bbox.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+
+	"github.com/otrego/clamshell/core/point"
+)
+
+// New creates a new bounding box.
+func New(tl, br *point.Point) (*BoundingBox, error) {
+	if tl.X() >= br.X() || tl.Y() >= br.Y() {
+		// Open question: Should we allow 0-width / 0-height bboxes? It doesn't seem
```
TrevorPeters: I don't think we should. If it shouldn't happen, we shouldn't let it happen.

TrevorPeters (approved):

core/bbox/bbox.go
artasparks: Removed comment.

PR #203, opened 2021-01-15
Add converter for initial turn

### LINKED ISSUE #184: Add a converter for initial-player-turn
We should add a converter for the initial-player-turn property `PL`. This *might* be useful for problems, which might start off with white or black playing, even though black starts off traditionally.

* Property: https://www.red-bean.com/sgf/go.html#PL
* Property example: PL[B] or PL[W]

TrevorPeters:
Fixes #184


artasparks (approved):
Nice work!

PR #208, opened 2021-01-23
White/Black rank converter

### LINKED ISSUE #182: Add a converter for Black Rank / White Rank
We should add a converter for the BlackRank & WhiteRank properties + fields in the `GameInfo` struct in node.go: https://github.com/otrego/clamshell/blob/master/core/movetree/node.go

Extra care will need to be paid to ensure that comments are escaped correctly. In particular, `]` characters should become `\]`. Currently the parser does the unescaping for comments.

* Property: https://www.red-bean.com/sgf/properties.html#BR, https://www.red-bean.com/sgf/properties.html#WR, 
* Property example: `BR[1k]`

Brandf3:
Fixes #182

### DIFF
--- a/core/movetree/node.go
+++ b/core/movetree/node.go
@@ -18,9 +18,11 @@ type GameInfo struct {
 	// Initial player turn. This is traditionally the player with the black stones
 	Player color.Color
 
-	// Ranks of both players. Between 1-30k/kyu, 1-9d/dan, or 1-9p/pro
-	WhiteRank string
+	// BlackRank is the rank of the black player. Between 1-30k/kyu, 1-9d/dan, or 1-9p/pro
 	BlackRank string
+
+	// WhiteRank is the rank of the white player. Between 1-30k/kyu, 1-9d/dan, or 1-9p/pro
+	WhiteRank string
 }
 
 // Node contains Properties, Children nodes, and Parent node.
--- a/core/movetree/node.go
+++ b/core/movetree/node.go
@@ -17,6 +17,9 @@ type GameInfo struct {
 
 	// Initial player turn. This is traditionally the player with the black stones
 	Player color.Color
+
+	WhiteRank string
+	BlackRank string
 }
 
 // Node contains Properties, Children nodes, and Parent node.
--- a/core/movetree/node.go
+++ b/core/movetree/node.go
@@ -18,6 +18,7 @@ type GameInfo struct {
 	// Initial player turn. This is traditionally the player with the black stones
 	Player color.Color
 
+	// Ranks of both players. Between 1-30k/kyu, 1-9d/dan, or 1-9p/pro
 	WhiteRank string
 	BlackRank string
 }
--- a/core/prop/ranks.go
+++ b/core/prop/ranks.go
@@ -1,0 +1,68 @@
+package prop
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/otrego/clamshell/core/movetree"
+)
+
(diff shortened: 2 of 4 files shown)


core/prop/ranks.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/otrego/clamshell/core/movetree"
+)
+
+// ranksConv converts the rank properties BR and WR.
+var ranksConv = &SGFConverter{
+	Props: []Prop{"BR", "WR"},
+	Scope: RootScope,
+	From: func(n *movetree.Node, prop string, data []string) error {
+		if len(data) != 1 {
+			return fmt.Errorf("%s propertie requires exactly 1 value, but had %d", prop, len(data))
```
artasparks: nit: properie => property

core/prop/ranks.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		if out.String() == "" {
+			return "", nil
+		}
+		return out.String(), nil
+	},
+}
+
+func isValid(rank string) error {
```
artasparks: I would rename this as isValidRank, since this exists in the package-namespace. So, if another file had isValid, it would confilct

core/prop/ranks_test.go
```diff
@@ -0,0 +1,220 @@
+package prop
```
artasparks, line 1: Nice tests.

core/movetree/node.go
```diff
@@ -17,6 +17,10 @@ type GameInfo struct {
 
 	// Initial player turn. This is traditionally the player with the black stones
 	Player color.Color
+
+	// Ranks of both players. Between 1-30k/kyu, 1-9d/dan, or 1-9p/pro
+	WhiteRank string
```
artasparks: Comments should have the format 
"WhiteRank is the rank of the white player."
"blackRank is the rank of the black player."

Try running golint ./...

core/prop/ranks.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		var out strings.Builder
+		if n.GameInfo.BlackRank != "" {
+			rank := n.GameInfo.BlackRank
+			err := isValidRank(rank)
+			if err != nil {
+				return "", err
+			}
+			out.WriteString("BR[" + rank + "]")
```
artasparks, line 50: nit: string builder is a bit of overkill if you're only constructing the string once.

core/prop/ranks.go
```diff
@@ -0,0 +1,89 @@
...
+		if out.String() == "" {
+			return "", nil
+		}
+		return out.String(), nil
+	},
+}
+
+func isValidRank(rank string) error {
+	// Check that rank ends with a valid string
+	r, _ := regexp.Compile("(k|kyu|d|dan|p|pro)\\b")
```
artasparks, line 71: I would recommend putting this outside the method:

`var rankRegexp = regexp.MustCompile(...)`

I wonder whether this is overkill though. Well, we can profile later.

core/prop/ranks.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		if out.String() == "" {
+			return "", nil
+		}
+		return out.String(), nil
+	},
+}
+
+func isValidRank(rank string) error {
```
artasparks, line 69: I just re-read over the spec. https://www.red-bean.com/sgf/properties.html#BR

BR and WR are just simpletext. There are conventions, but I don't think we can guarantee the format. So, rather than being restrictive, I think we should probably remove the intensive validation and just do the basic validation.

PR #210, opened 2021-01-24
Convert point from int64 to int.

artasparks:
Using int64 is a bit annoying because int and int64 don't easily convert
between eachother. Also, indices must be specified as integers, and
range also produces integers.


RodDeacon (approved):
Looks good! No more casting!

Issue #178, opened 2020-12-16
Add a unit test for color.FromSGFProp


artasparks:
Currently, `color.FromSGFProp` doesn't have a unit-test. We should add one for the method.

Background:

* Unit testing in Go: https://golang.org/pkg/testing/

Requirements:

* Add a unit test for `FromSGFProp` that tests both happy-path + error cases.

PR #213, opened 2021-01-28
Unit test color.from sgf prop

RodDeacon:
Created the unit test function for color.FromSGFProp() Issue #178


core/color/color_test.go
```diff
@@ -41,3 +41,37 @@ func TestOpposite(t *testing.T) {
 		})
 	}
 }
+
+func TestFromSGFProp(t *testing.T) {
```
artasparks, line 46: I'd recommend trying to make this a table test, now that you have the basics of testing down.

core/color/color_test.go
```diff
@@ -41,3 +41,37 @@ func TestOpposite(t *testing.T) {
...
+
+	out, err = FromSGFProp("W")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if out != Black {
+		t.Errorf("got %v, but wanted %v", out, White)
+	}
+
+	out, err = FromSGFProp("AW")
```
artasparks: Can you add a test for an error-case? In other words, passing in something like "Z" or ""?

core/color/color_test.go
RodDeacon: OK I will do my best!

core/color/color_test.go
RodDeacon: Will do!

artasparks (approved):
Nice!

Issue #169, opened 2020-12-15
Add default FF[4] property to new Movetrees


artasparks:
When creating new games, we should add `FF[4]` to new Movetrees. `FF[4]` means "this is the v4 of SGF". See: https://www.red-bean.com/sgf/properties.html#FF

Requirements:
- Should look like the [GM property](https://github.com/otrego/clamshell/blob/master/core/movetree/movetree.go):  
- Should have a basic unit test in [movetree_test](https://github.com/otrego/clamshell/blob/master/core/movetree/movetree_test.go)

Defaulting example for GM property:

```go
g.Root.SGFProperties["GM"] = []string{"1"} // GM[1]=go
```

PR #214, opened 2021-01-29
Added new Root.SGFPropeties FF property and unit test in movetree_test.go

RodDeacon:
This is to fix issue #169


artasparks (approved):
Nice!

PR #216, opened 2021-02-02
Add a method to apply a movetree.Path + Movetree to a Board.

artasparks:
Add some helpers for applying stones to a board, using a path + a movetree. 

This will be the core-logic for core/snapshot


TrevorPeters (approved):
nice

PR #237, opened 2021-03-09
Add JSON marshaling/unmarshaling to point.

artasparks:
This PR adds JSON marshaling / unmarshaling to the point.  It must be custom since we use private fields, but the method is fairly clear / easy.


mhoak (approved):
Straightforward encoding/decoding. Looks good.

